PR #60, opened 2022-03-04
Add license info. And delete old files no longer needed.

rschwant:


zmoon:
> I think the license needs to be in the main branch for the badge to work

For posterity, it seems the repo also had to be public so the service for the badge could detect it

PR #64, opened 2022-03-04
Add Read the Docs config file

zmoon:
specifying to use the `melodies-monet-docs` conda env


zmoon:
I guess with the branch protection rules we should go to `develop` first so that is there

rschwant (approved):
Looks good.

zmoon:
Ok, let's try it. I enabled building the `develop` branch in RtD so we will be able to check things.

PR #68, opened 2022-03-04
Update machine-specific-install.rst

lkemmons:
Corrected out-of-date link to CISL instructions.


zmoon (approved):
too bad the old one doesn't redirect

lkemmons:
Yes, they rearranged the whole CISL website and nothing is redirected...
You'd think the computing lab would do a better job with that!

PR #70, opened 2022-03-08
Update the install docs

lkemmons:
minor clarifying edits.


docs/appendix/machine-specific-install.rst
```diff
@@ -1,8 +1,8 @@
 Machine-specific Install
 ========================
 
-NCAR HPC Cheyenne
------------------
+NCAR HPC Cheyenne or Casper
```
zmoon, line 4: Changing this section title broke a reference to it [elsewhere](https://github.com/NOAA-CSL/MELODIES-MONET/blame/b9aee5c0fd3f9d2212306e663169c9db40e27655/docs/tutorial/installation.rst#L29), that's why the check failed. Casper is mentioned below, so maybe it would be ok to leave the heading?

docs/appendix/machine-specific-install.rst
```diff
@@ -76,7 +76,8 @@ dependencies.
     environment is still activated, get the most recent branches of MONET and 
     MONETIO using GitHub, and link them with conda. This is done because MONET
     and MONETIO are still in active development. Create a 'monet-base' folder
-    (e.g. in your work location on cheyenne).
+    (e.g. in your /glade/work directory on cheyenne; you can put the monet and
```
zmoon, line 79:
```suggestion
    (e.g. in your ``/glade/work`` directory, accessible on both Cheyenne and Casper; you can put the monet and 
```

docs/appendix/machine-specific-install.rst
```diff
@@ -94,9 +95,9 @@ dependencies.
...
 **Step 3: Clone the MELODIES-MONET GitHub repository** ::
 
-    $ git clone git@github.com:NOAA-CSL/MELODIES-MONET.git
+    $ git clone https://github.com/NOAA-CSL/MELODIES-MONET.git
 
-    End step. At the end of working with MELODIES MONET, deactivate the 
+    At the end of working with MELODIES MONET, deactivate the 
     conda environment::
 
     $ conda deactivate
```
zmoon, line 103: I think we could just remove this. The Hera section doesn't have it, and it isn't related to "Clone the MELODIES-MONET GitHub repository", which it is nested under.

image

lkemmons:
Thanks!  That sounds fine.
Sorry to be barging in and breaking things...

docs/appendix/machine-specific-install.rst
rschwant: This title is mentioned in docs/tutoriall/installation.rst. We can add Casper we should just update it in both places.

lkemmons:
I think it is fine to leave the title as it was.  As Zach said, Casper is
mentioned in the text below.

lkemmons:
Yes, that might be better.  Thanks!

zmoon:
@lkemmons would you be able to check https://github.com/NOAA-CSL/MELODIES-MONET/pull/70/files and make sure I got all the changes you wanted?

lkemmons:
That looks fine.
Do we have instructions for non-developers - people who would just like to
use the existing code and explore what it can do?

zmoon:
> Do we have instructions for non-developers - people who would just like to use the existing code and explore what it can do?

Hmm, I think not exactly yet. Currently they could just skip the parts about cloning the MONET and MONETIO repos but would still need to clone *this* repository. I could add a note to that effect in the main install doc.

At some point I imagine we will get this package on conda-forge so that only
```
conda install -c conda-forge melodies-monet
```
would be required.

@rschwant ?

rrbuchholz:
There is a place for general instructions in the user guide:
https://melodies-monet.readthedocs.io/en/develop/tutorial/installation.html#general-instructions
They mirror the developer instructions a lot right now, but will change in future.

zmoon:
> They mirror the developer instructions a lot right now

Yeah that's what I was referring to, I realize that the section does exist

zmoon:
> For now, we need people to still download the develop branches of monet and monetio, so the general instructions in the readthedocs are the only way to run the package.

Currently the conda-forge versions of monet/monetio are up to date with develop (I did releases yesterday), but that won't always be the case, so I agree better to leave it how it is for now.

rschwant (approved):
This is a great compromise and very clear for new users. Thanks!

PR #76, opened 2022-03-10
Update installation.rst

lkemmons:
Added a couple clarifying sentences at start of general instructions for non-developers.

### DIFF
--- a/docs/tutorial/installation.rst
+++ b/docs/tutorial/installation.rst
@@ -21,6 +21,11 @@ of all the dependencies are available via the GitHub:
 General Instructions
 --------------------
 
+The instructions below allow users to contribute to the development of
+MELODIES MONET (which is encouraged!). If you prefer to be a casual user, you may
+clone the monet, monetio and MELODIES-MONET repositories using HTTPS instead 
+of ssh [#clone]_.
+
 Setting up MELODIES MONET on HPC machines can be complicated. To help users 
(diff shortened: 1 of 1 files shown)


rschwant:
What about something like that below. Right now we don't explain here how to setup a fork etc., so the reason for using SSH for development is more later on. But this explains the different options for users less familiar with GitHub. Technically, you can really use either approach and everyone will have their preference. We should add this to the develop branch too as I want to keep the user guide consistent. To make it more seamless, Louisa you could push this first to develop and then from develop to main.

@zmoon @lkemmons 

The instructions below are for cloning a repository using SSH, which is recommended, but requires adding an SSH key on your current machine. If you prefer, you can also clone the monet, monetio, and MELODIES-MONET repositories using HTTPS [#clone]_.

rschwant:
> 

There was a lot of confusion at NCAR during the practice tutorial because people did not see the footnote. We could condense it even more.

zmoon:
condense?

rschwant:
The instructions below are for cloning a repository using SSH. If you prefer, you can also clone the monet, monetio, and MELODIES-MONET repositories using HTTPS [#clone]_.

rschwant:
Yes that works. I like the important one. Let's add that to the develop and main branches.

lkemmons:
Looks great!  Thanks!

zmoon:
In favor of #77

PR #77, opened 2022-03-10
Add more clone note to main install doc

zmoon:
Add a more-obvious note about the cloning to the main installation page of the docs.

Based on discussion in #76


rschwant (approved):
Thanks for updating this

PR #81, opened 2022-04-12
Update install docs

### LINKED ISSUE #79: Hera dev instructions can't use PyPI
The current recommended dev install method [^1]
```
pip install --no-deps --editable .
```
doesn't work[^2] because the build process tries to download a fresh setuptools from PyPI, but Hera doesn't allow downloading from there.

An alternative (which has been tested to work) is
```
python setup.py develop --no-deps
```

[^1]: Currently recommended for monet/monetio
[^2]: This was reported by Megan Bela.

zmoon:
With #80 merged, we can feel pretty good about people `pip install`ing the package. This means that we support and can recommend a simpler `git`-less install for users. I have moved the development install instructions to the dev guide page.

Closes #79


zmoon:
@rschwant let me know what you think we should do about the Hera/Cheyenne instructions. I am thinking many parts could be eliminated, just referencing the other install instructions, but we can leave some of the notes that are specific to those machines.

rschwant:
I would be okay with simplifying these instructions for Hera now. Now the only difference with the main instructions is you need to link the cartopy files, so this entire section could be made much simpler. @rrbuchholz how do you feel about simplifying the instructions for Cheyenne now that the process for loading MONET and MONETIO is easier. Zach I am wondering if in the main instructions we want to just move to using http cloning rather than ssh cloning. I keep getting emails from people learning who have not set up an ssh key. To make it simpler we can leave the main instructions as http and the develop instructions as ssh. Developers would likely understand the difference, but people who are very new to Python and GitHub do not. Does this work for you?

zmoon:
@rschwant for the main instructions I am suggesting not to do any cloning. See them [here](https://github.com/zmoon/MELODIES-MONET/blob/install-docs/docs/tutorial/installation.rst) and the developer's guide ones [here](https://github.com/zmoon/MELODIES-MONET/blob/install-docs/docs/develop/developers_guide.rst#setting-up-your-development-environment).

rschwant:
Great that makes sense. The how_to_run.rst and all of the example jupyter notebooks and run scripts still say the following:
import sys; sys.path.append("../../")
from melodies_monet import driver

With this pip install, you can just add the second line correct:
from melodies_monet import driver

Do you mind updating all of these lines for consistency? Perhaps, something like the following
#If you have cloned the MELODIES-MONET repository instead of using pip install, uncomment line below and point to your cloned repository
#import sys; sys.path.append("../../")
from melodies_monet import driver

zmoon:
> With this pip install, you can just add the second line correct:
> `from melodies_monet import driver`

@rschwant thanks for pointing that out, I will fix those. `import sys; sys.path.append("../../")` isn't needed for the dev install either since we are `pip install`ing MM there too, just the local one instead of from the GitHub url.

rschwant:
Great, then this is even simpler. Thanks for updating these other files too.

zmoon:
Current CI failure is because it is (currently) unable to reach the tutorial data on the CSL website

rschwant:
> Current CI failure is because it is (currently) unable to reach the tutorial data on the CSL website, and the linkcheck showed that other NOAA domains are down as well.

Yes, the systems will be back up by this afternoon.

docs/tutorial/installation.rst
```diff
@@ -1,84 +1,49 @@
...
-- All required dependencies for `MONET <https://monet-arl.readthedocs.io/en/stable/installing.html>`__
-- All required dependencies for `MONETIO <https://monetio.readthedocs.io/en/stable/installing.html>`__
+Required dependencies [#yaml]_
+------------------------------
 
-Conda YAML files
-~~~~~~~~~~~~~~~~
-Examples of conda configuration environment.yml files that include a record
-of all the dependencies are available via the GitHub:
+- Python 3.7+ (3.9 recommended)
```
zmoon, line 7:
```suggestion
- Python 3.6+ (3.9 recommended)
```

rrbuchholz (approved):
Looks good to me.

zmoon:
Thanks @rrbuchholz

docs/tutorial/how_to_run.rst
```diff
@@ -160,3 +158,7 @@ are in the ``examples/submit_jobs`` folder of the code on GitHub.
 
    * Submit the submit_hera.sh script (e.g., on NOAA hera: sbatch submit_hera.sh)	 
 
+
+.. [#install] This should work from any directory as long is MELODIES MONET is installed
```
rschwant, line 162:
```suggestion
.. [#install] This should work from any directory as long as MELODIES MONET is installed
```

rschwant (approved):
Looks great! I added one typo comment. At some point, we should consider updating the hera.yml and cheyenne.yml files as they may be slightly outdated, but we are not recommending anyone use them anymore, so it's less critical for right now. Let's plan to make sure these are updated whenever we have a version release.

zmoon:
Thanks @rschwant , sounds good.

PR #84, opened 2022-05-16
Develop reg2

jianheRCM:
Hi all,

I have included some updates on the regulatory options. Could you please help review it?
I created a new branch develop_reg2 because somehow I was not able to push my changes to develop_reg.

Thanks,


melodies_monet/driver.py
```diff
@@ -1066,16 +1139,48 @@ def stats(self):
...
+                                print('No valid data for '+obsvar+'_reg')
+                                cal_reg = False
+                            else:
+                                # Drop NaNs for model and observations in all cases.
+                                pairdf2 = pairdf_reg.reset_index().dropna(subset=[modvar+'_reg', obsvar+'_reg'])
+                                outname = "{}.{}.{}.{}.{}.{}".format('stats', obsvar+'_reg', domain_type, domain_name, startdatename, enddatename)
+                                title = obs_plot_dict['ylabel2_plot'] + ': ' + domain_type + ' ' + domain_name
+
+                            del(df1)
+                            del(df2)
```
zmoon, line 1165:
```suggestion
                            del df1, df2
```

melodies_monet/driver.py
```diff
@@ -1066,16 +1139,48 @@ def stats(self):
...
+                                pairdf_reg = splots.make_8hr_regulatory(df2,[obsvar,modvar]).rename(index=str,columns={obsvar+'_y':obsvar+'_reg',modvar+'_y':modvar+'_reg'})
+                            else:
+                                print('Warning: no regulatory caculations found for ' + obsvar)
+
+                            if len(pairdf_reg[obsvar+'_reg']) == 0:
+                                print('No valid data for '+obsvar+'_reg')
+                                cal_reg = False
+                            else:
+                                # Drop NaNs for model and observations in all cases.
+                                pairdf2 = pairdf_reg.reset_index().dropna(subset=[modvar+'_reg', obsvar+'_reg'])
```
zmoon, line 1160: `reset_index()` after `dropna()` instead?

melodies_monet/plots/surfplots.py
```diff
@@ -138,7 +157,66 @@ def map_projection(f):
...
+    -------
+    UTC offset in hour
+
+    """
+
+    from datetime import datetime
+    import pytz
+    from datetime import datetime, timezone
+
+    tf = TimezoneFinder()
```
zmoon, line 178: Should you be importing this within this function? Since you commented the module-level import?

melodies_monet/plots/surfplots.py
```diff
@@ -138,7 +157,66 @@ def map_projection(f):
...
+        else:
+            timezone_str = 'America/Los_Angeles'
+
+        tz = pytz.timezone(timezone_str)
+        d=datetime.utcnow()
+        uos = tz.utcoffset(d, is_dst=False)
+        utchour = uos.seconds/60.0/60.0
+        utcday = uos.days
+
+    elif timezone_str.startswith(('Etc','GMT')):
```
zmoon, line 197:
```suggestion
    elif timezone_str.startswith({'Etc', 'GMT'}):
```

melodies_monet/plots/surfplots.py
```diff
@@ -626,13 +756,17 @@ def make_spatial_overlay(df, vmodel, column_o=None, label_o=None, column_m=None,
...
+def calculate_boxplot(df, df_reg, regulatory=None, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
     """Combines data into acceptable format for box-plot
     
     Parameters
     ----------
     df : dataframe
         Model/obs pair object
+    df_reg: dataframe
+        model/obs pair data to plot
+    regulatory : boolean
```
zmoon, line 768:
```suggestion
    regulatory : bool
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        Title of colorbar axis
+    vdiff : real number
+        Min and max value to use on colorbar axis
+    outname : str
+        file location and name of plot (do not include .png)
+    domain_type : str
+        Domain type specified in input yaml file
+    domain_name : str
+        Domain name specified in input yaml file
+    fig_dict : dictionary
```
zmoon, line 940:
```suggestion
    fig_dict : dict
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        Dictionary containing information about figure
+    text_dict : dictionary
```
zmoon, line 942:
```suggestion
    text_dict : dict
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        Dictionary containing information about text
+    debug : boolean
```
zmoon, line 944:
```suggestion
    debug : bool
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+    label_o : str
+        Name of observation variable to use in plot title 
+    column_m : str
+        Column label of model variable to plot
+    label_m : str
+        Name of model variable to use in plot title
+    ylabel : str
+        Title of colorbar axis
+    vdiff : real number
+        Min and max value to use on colorbar axis
```
zmoon, line 933: As it is, this is a little confusing. Maybe calling it `vmax` would be better. And saying that `-vmax` will be used as `vmin` (if that's what is going on).

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        cax = f.get_axes()[1]
+        #get the position of the plot axis and use this to rescale nicely the color bar to the height of the plot.
+        position_m = model_ax.get_position()
+        position_c = cax.get_position()
+        cax.set_position([position_c.x0, position_m.y0, position_c.x1 - position_c.x0, (position_m.y1-position_m.y0)*1.1])
+        cax.set_ylabel(ylabel2+ ' exceedance days',fontweight='bold',**text_kwargs)
+        cax.tick_params(labelsize=text_kwargs['fontsize']*0.8,length=10.0,width=2.0,grid_linewidth=2.0)
+
+        #plt.tight_layout(pad=0)
+        savefig(outname + '_exceendance.png', loc=4, logo_height=120)
```
zmoon, line 1046:
```suggestion
        savefig(outname + '_exceedance.png', loc=4, logo_height=120)
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+    text_dict : dictionary
+        Dictionary containing information about text
+    debug : boolean
+        Whether to plot interactively (True) or not (False). Flag for 
+        submitting jobs to supercomputer turn off interactive mode.
+        
+    Returns
+    -------
+    plot 
+        surface bias plot
```
zmoon, line 951: This refers to the return value of the function, i.e. if you do `a = make_spatial_exceedance(...)`, what is `a`? It doesn't look you have specified to return anything, so this function just returns `None`, in which case you can document that or just leave off the "Returns" section of the docstring.

melodies_monet/plots/surfplots.py
```diff
@@ -138,7 +157,66 @@ def map_projection(f):
...
+        utchour = uos.seconds/60.0/60.0
+        utcday = uos.days
+
+    if utcday < 0:
+       utchour = (24-utchour)*(-1) # Local - UTC
+
+    return utchour
+
+
+def make_spatial_bias(df, df_reg, regulatory=None, column_o=None, label_o=None, column_m=None,
```
zmoon, line 219:
```suggestion
def make_spatial_bias(df, df_reg, regulatory=False, column_o=None, label_o=None, column_m=None, 
```

melodies_monet/plots/surfplots.py
```diff
@@ -249,7 +340,7 @@ def make_spatial_bias(df, column_o=None, label_o=None, column_m=None,
     #plt.tight_layout(pad=0)
     savefig(outname + '.png', loc=4, logo_height=120)
     
-def make_timeseries(df, column=None, label=None, ax=None, avg_window=None, ylabel=None,
+def make_timeseries(df, df_reg, regulatory=None, column=None, label=None, ax=None, avg_window=None, ylabel=None,
```
zmoon, line 343:
```suggestion
def make_timeseries(df, df_reg, regulatory=False, column=None, label=None, ax=None, avg_window=None, ylabel=None,
```

melodies_monet/plots/surfplots.py
```diff
@@ -358,7 +461,7 @@ def make_timeseries(df, column=None, label=None, ax=None, avg_window=None, ylabe
             ax.set_title(domain_name,fontweight='bold',**text_kwargs)
     return ax
     
-def make_taylor(df, column_o=None, label_o='Obs', column_m=None, label_m='Model', 
+def make_taylor(df, df_reg, regulatory=None, column_o=None, label_o='Obs', column_m=None, label_m='Model',
```
zmoon, line 464:
```suggestion
def make_taylor(df, df_reg, regulatory=False, column_o=None, label_o='Obs', column_m=None, label_m='Model', 
```

melodies_monet/plots/surfplots.py
```diff
@@ -626,13 +756,17 @@ def make_spatial_overlay(df, vmodel, column_o=None, label_o=None, column_m=None,
     savefig(outname + '.png', loc=4, logo_height=100, dpi=150)
     return ax
     
-def calculate_boxplot(df, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
+def calculate_boxplot(df, df_reg, regulatory=None, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
```
zmoon, line 759:
```suggestion
def calculate_boxplot(df, df_reg, regulatory=False, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
```

zmoon (commented):
Great to see this coming in @jianheRCM . I've taken an initial pretty quick look.

Some general comments:
* would be good to clean up in some places, e.g. reinstate or remove commented code (Git will have a record of this, so if you need to go back at some point you can)
* I think it would be nicer for `df_reg` to be an optional argument instead of required. For example, you could give it a default of `None` (and then in function body check for `None` and convert to empty dataframe if you want)
* we will need to update install docs and conda env specs to account for the new dependencies you introduce

examples/yaml/control_rrfs_cmaq_airnow_norm.yaml
```diff
@@ -0,0 +1,226 @@
...
+        nan_value: -1.0 # Set this value to NaN
+        #The obs_min, obs_max, and nan_values are set to NaN first and then the unit conversion is applied.
+        ylabel_plot: 'PM2.5 (ug/m3)' #Optional to set ylabel so can include units and/or instr etc.
+        ty_scale: 3.0 #Opt
+        vmin_plot: 0.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 22.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 15.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 23 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: False #Opt compute regulatory functions
+        ylabel2_plot: 'PM2.5_24hr (ug/m3)'
```
rschwant, line 88: Can we change here and in the code this to ylabel_reg_plot? Just so that it is clearer what it is for.

rschwant, line 88: Can we change here and in the code this to ylabel_reg_plot? Just so that it is clearer what it is for.

melodies_monet/driver.py
```diff
@@ -720,8 +720,8 @@ def plotting(self):
 
                         # Determine outname
                         outname = "{}.{}.{}.{}.{}.{}.{}".format(grp, plot_type, obsvar, startdatename, enddatename, domain_type, domain_name)
-                        if self.output_dir is not None:
-                            outname = self.output_dir + '/' + outname  # Extra / just in case.
+                        #if self.output_dir is not None:
+                        #    outname = self.output_dir + '/' + outname  # Extra / just in case.
```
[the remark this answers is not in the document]
rschwant: Looking at the code, I think this whole section under #Determine outname can be removed. 
So this line
outname = "{}.{}.{}.{}.{}.{}.{}".format(grp, plot_type, obsvar, startdatename, enddatename, domain_type, domain_name)
and the lines you commented Jian.

Jian it looks like you added this now with in the regulatory if else statement, which works for me. We don't need it twice.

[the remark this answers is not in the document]
rschwant: Looking at the code, I think this whole section under #Determine outname can be removed. 
So this line
outname = "{}.{}.{}.{}.{}.{}.{}".format(grp, plot_type, obsvar, startdatename, enddatename, domain_type, domain_name)
and the lines you commented Jian.

Jian it looks like you added this now with in the regulatory if else statement, which works for me. We don't need it twice.

melodies_monet/driver.py
```diff
@@ -734,6 +734,42 @@ def plotting(self):
...
+                            else:
+                                print('Warning: no regulatory caculations found for ' + obsvar) 
+
+                            if len(pairdf_reg[obsvar+'_reg']) == 0:
+                                print('No valid data for '+obsvar+'_reg')
+                                cal_reg = False
+                            else:
+                                #print('Paired dataframe with regulatory: ', obsvar, pairdf_reg)
+                                outname = "{}.{}.{}.{}.{}.{}.{}".format(grp, plot_type, obsvar+'_reg', startdatename, enddatename, domain_type, domain_name)
+                                use_ylabel = obs_plot_dict['ylabel2_plot']
```
rschwant, line 762: See above, can we change the name to be more specific like ylabel_reg_plot. Also can we change the code to be similar to that above for the ylabel_plot where if it is not provided you set use_ylabel = none. That way if people do not want to provide a name in the yaml file that's okay it will default to the name of the obsvar name.

rschwant, line 762: See above, can we change the name to be more specific like ylabel_reg_plot. Also can we change the code to be similar to that above for the ylabel_plot where if it is not provided you set use_ylabel = none. That way if people do not want to provide a name in the yaml file that's okay it will default to the name of the obsvar name.

melodies_monet/driver.py
```diff
@@ -734,6 +734,42 @@ def plotting(self):
...
+
+                            del(df1)
+                            del(df2)
```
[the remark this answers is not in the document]
rschwant: If this ends up saving memory to delete these extra variables, I'm okay leaving it in.

melodies_monet/driver.py
```diff
@@ -937,25 +1000,27 @@ def plotting(self):
...
-                                proj=proj,
-                                outname=outname,
-                                domain_type=domain_type,
-                                domain_name=domain_name,
-                                fig_dict=fig_dict,
-                                text_dict=text_dict,
-                                debug=self.debug
-                            )
+                            # JianHe: only make overplay plots for non-regulatory variables for now
+                            if not cal_reg:
```
rschwant, line 1004: Can you add a message here like

else:
     print('Warning: Spatial overlay plots are not available yet for regulatory metrics.')

melodies_monet/driver.py
```diff
@@ -1017,8 +1090,8 @@ def stats(self):
                 df_o_d = pd.DataFrame()
                 # Determine outname
                 outname = "{}.{}.{}.{}.{}.{}".format('stats', obsvar, domain_type, domain_name, startdatename, enddatename)
-                if self.output_dir is not None:
-                    outname = self.output_dir + '/' + outname  # Extra / just in case.
+                #if self.output_dir is not None:
+                #    outname = self.output_dir + '/' + outname  # Extra / just in case.
```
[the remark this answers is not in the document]
rschwant: Let's update this like we did above in the plotting routine.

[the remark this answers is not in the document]
rschwant: Let's update this like we did above in the plotting routine.

melodies_monet/driver.py
```diff
@@ -1066,16 +1139,48 @@ def stats(self):
...
+                                print('Warning: no regulatory caculations found for ' + obsvar)
+
+                            if len(pairdf_reg[obsvar+'_reg']) == 0:
+                                print('No valid data for '+obsvar+'_reg')
+                                cal_reg = False
+                            else:
+                                # Drop NaNs for model and observations in all cases.
+                                pairdf2 = pairdf_reg.reset_index().dropna(subset=[modvar+'_reg', obsvar+'_reg'])
+                                outname = "{}.{}.{}.{}.{}.{}".format('stats', obsvar+'_reg', domain_type, domain_name, startdatename, enddatename)
+                                title = obs_plot_dict['ylabel2_plot'] + ': ' + domain_type + ' ' + domain_name
```
rschwant, line 1162: Like the plotting routines, can we make this name more specific ylabel_reg_plot and also optional. You could add this even to the if statement above on line 1112. Such that if cal_reg = True you use a different label.

rschwant, line 1162: Like the plotting routines, can we make this name more specific ylabel_reg_plot and also optional. You could add this even to the if statement above on line 1112. Such that if cal_reg = True you use a different label.

melodies_monet/plots/surfplots.py
```diff
@@ -21,6 +21,9 @@
 from monet.util.tools import get_epa_region_bounds as get_epa_bounds 
 import math
 from ..plots import savefig
+import datetime, pytz
```
rschwant, line 24: On line 10, since you don't use the calc_8hr_rolling_max, calc_24hr_ave from MONET anymore. Should we delete this line?

melodies_monet/plots/surfplots.py
rschwant: Yes, let's move this out of this function and up to the top of the code instead. Maybe we want to call this function for other uses.

rschwant: Yes, let's move this out of this function and up to the top of the code instead. Maybe we want to call this function for other uses.

zmoon: @rschwant not sure what you mean. This line is within a function. My comment was about `TimezoneFinder` not being imported anywhere, since

https://github.com/NOAA-CSL/MELODIES-MONET/blob/a873dd268c475b67c030007068c8b852da22348d/melodies_monet/plots/surfplots.py#L25

is commented

zmoon:
> * I think it would be nicer for `df_reg` to be an optional argument instead of required. For example, you could give it a default of `None` (and then in function body check for `None` and convert to empty dataframe if you want)

@rschwant did you agree with this?

melodies_monet/plots/surfplots.py
rschwant: Oh I misunderstood. I see what you mean. I'm open if we don't want to update the dependencies of MELODIES-MONET to move all of these packages to only import if you call this function. I'm thinking this will not be the default function because it will be very slow and we do not yet call it in the driver. Users can then load this package if they need it separately. Do you all agree?

melodies_monet/plots/surfplots.py
```diff
@@ -150,6 +228,10 @@ def make_spatial_bias(df, column_o=None, label_o=None, column_m=None,
     ----------
     df : dataframe
         model/obs pair data to plot
+    df_reg: dataframe
+        model/obs pair data to plot
```
rschwant, line 232:
```suggestion
        model/obs pair of regulatory data to plot
```

melodies_monet/plots/surfplots.py
```diff
@@ -259,6 +350,10 @@ def make_timeseries(df, column=None, label=None, ax=None, avg_window=None, ylabe
...
```
rschwant, line 354:
```suggestion
        model/obs pair of regulatory data to plot
```

melodies_monet/plots/surfplots.py
```diff
@@ -369,6 +472,10 @@ def make_taylor(df, column_o=None, label_o='Obs', column_m=None, label_m='Model'
...
```
rschwant, line 476:
```suggestion
        model/obs pair of regulatory data to plot
```

melodies_monet/plots/surfplots.py
```diff
@@ -626,13 +756,17 @@ def make_spatial_overlay(df, vmodel, column_o=None, label_o=None, column_m=None,
...
-def calculate_boxplot(df, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
+def calculate_boxplot(df, df_reg, regulatory=None, column=None, label=None, plot_dict=None, comb_bx = None, label_bx = None):
     """Combines data into acceptable format for box-plot
     
     Parameters
     ----------
     df : dataframe
         Model/obs pair object
+    df_reg: dataframe
+        model/obs pair data to plot
```
rschwant, line 767:
```suggestion
        model/obs pair of regulatory data to plot
```

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+                      outname = 'plot',
+                      domain_type=None, domain_name=None, fig_dict=None,
+                      text_dict=None,debug=False):
+
+    """Creates surface spatial bias plot. 
+    
+    Parameters
+    ----------
+    df : dataframe
+        model/obs pair data to plot
```
rschwant, line 921:
```suggestion
        model/obs pair of regulatory data to plot
```

melodies_monet/driver.py
```diff
@@ -898,7 +944,24 @@ def plotting(self):
                                 text_dict=text_dict,
                                 debug=self.debug
                             )
+                            if cal_reg:
+                                outname = "{}.{}".format(outname, p_label)
+                                splots.make_spatial_exceedance(
```
rschwant, line 949: Do you want to make the spatial exceedences plot, a plot_type instead? You could specify in the yaml file and readthedocs that the plot is only for regulatory calculations and add an if statement to only make the plot for regulatory calculations. This would make things more consistent, but we can also update this later if you prefer.

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +906,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        ylabel = column_o
+
+    #Take the mean for each siteid
+    #df_mean=df.groupby(['siteid'],as_index=False).mean()
+
+    #df_reg = pd.DataFrame()
+    print(df)
+    # calculate exceedance
+    if column_o == 'OZONE_reg':
+        # JianHe: we can get mean, or 95th, 99th percentile
```
rschwant, line 984: If you make the spatial_exceedance a plot_type instead of an add on to the spatial bias plot. You can add options in the yaml file calculating the mean or percentiles here if needed.

If this is a separate plotting group it also enables more flexibility for the axis. You can make it possible to use a different default max or even add one to the list like vmax_ct or something. And in the yaml file if you set set_axis: False then you can configure it so that it just uses a default vmax value, which may be the best for plots like this.

rschwant:
This looks great Jian! @jianheRCM This is all very close. I added many thoughts, but they are all minor. I have two questions to think about NaNs. I realized that you dropped the NaNs after the regulator calculations in the stats, but not the plots. Do you think we should drop the NaNs for the plots too? Did you think about this?

Another thought also with NaNs. We leave it up to the user to specify if they want to drop all values where the model is NaN or all the values where the obs or model are NaNs for the plotting. For the regulatory calculations, I'm thinking we should require that people select to drop the values where the obs or the model are NaNs and not leave this up to the user, so that these calculations are all done consistently. What do you all think? Or we could also add a warning to make sure this selection was intentional?

melodies_monet/plots/surfplots.py
zmoon: Yeah, I think the faster longitude-based method could be the default. And I think we could offload these options/details to MONET, but that can be after this PR. For now, the `from timezonefinder import TimezoneFinder` should be within the function so that only users who want reg need to install it.

rschwant:
> > * I think it would be nicer for `df_reg` to be an optional argument instead of required. For example, you could give it a default of `None` (and then in function body check for `None` and convert to empty dataframe if you want)
> 
> @rschwant what did you think about this?

That's a good point Zach. Yes, let's make the df_reg = None in these plots as default. Then you can always check if regulatory = True and df_reg is not None, plot regulatory option.

melodies_monet/driver.py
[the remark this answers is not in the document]
jianheRCM: I will delete the commented lines but keep the line within regulatory if/else statement

melodies_monet/driver.py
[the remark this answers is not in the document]
jianheRCM: I will delete these lines.

melodies_monet/driver.py
jianheRCM: I followed the previous statement:

e.g., pairdf = pairdf_all.reset_index().dropna(subset=[modvar, obsvar])

just to make it consistent...

melodies_monet/plots/surfplots.py
jianheRCM: I have a function get_utcoffset in the code in case in the future we want to calculate timezone more accurately. The function get_utcoffset will use below packages:
    from datetime import datetime
    import pytz
    from datetime import datetime, timezone
    from timezonefinder import TimezoneFinder

I will put TimezoneFinder within the function as above.

melodies_monet/plots/surfplots.py
jianheRCM: vmax will be set to be as vdiff like those in function make_spatial_bias.

examples/yaml/control_rrfs_cmaq_airnow_norm.yaml
jianheRCM: Good idea. Will update them.

melodies_monet/plots/surfplots.py
jianheRCM: If we do not need keep it as an option, or will not use them in the future, we can delete this line

jianheRCM:
>

jianheRCM:
> Great to see this coming in @jianheRCM . I've taken an initial pretty quick look.
> 
> Some general comments:
> 
> * would be good to clean up in some places, e.g. reinstate or remove commented code (Git will have a record of this, so if you need to go back at some point you can)
> * I think it would be nicer for `df_reg` to be an optional argument instead of required. For example, you could give it a default of `None` (and then in function body check for `None` and convert to empty dataframe if you want)
> * we will need to update install docs and conda env specs to account for the new dependencies you introduce

Thanks @zmoon for your suggestions. Set df_reg to be an optional argument is a good idea. I put both df and df_reg in the argument but only one of them will be used in the function depending on the boolean variable "regulatory".

jianheRCM:
> This looks great Jian! @jianheRCM This is all very close. I added many thoughts, but they are all minor. I have two questions to think about NaNs. I realized that you dropped the NaNs after the regulator calculations in the stats, but not the plots. Do you think we should drop the NaNs for the plots too? Did you think about this?
> 
> Another thought also with NaNs. We leave it up to the user to specify if they want to drop all values where the model is NaN or all the values where the obs or model are NaNs for the plotting. For the regulatory calculations, I'm thinking we should require that people select to drop the values where the obs or the model are NaNs and not leave this up to the user, so that these calculations are all done consistently. What do you all think? Or we could also add a warning to make sure this selection was intentional?

@rschwant  NaNs are actually dropped during the regulatory calculations. pairdf_reg is the output after NaNs dropped. So  we actually do not need drop NaNs again for the stats.

jianheRCM:
@zmoon @rschwant Thanks for your suggestions and comments. I have updated the code as suggested (cleanup, function arguments, etc). I also include plotting parameters specified for regulatory plots in the yaml file. If you can review the code again, that would be great.

melodies_monet/plots/surfplots.py
rschwant: I would just delete this line. And also delete the commented lines where you call these functions since we are not using them anymore it makes it easier for people to understand how the code works.

examples/yaml/control_rrfs_cmaq_airnow_reg.yaml
```diff
@@ -0,0 +1,246 @@
...
+        unit_scale_method: '*' #Opt Multiply = '*' , Add = '+', subtract = '-', divide = '/'
+        nan_value: -1.0 # Opt Set this value to NaN
+        ylabel_plot: 'Ozone (ppbv)'
+        vmin_plot: 15.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 55.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 20.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 21 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: True #Opt compute regulatory functions
+        ylabel_reg_plot: 'MDA8 O3 (ppbv)'
+        percentile_opt: 0.95 #Opt, 0.05, 0.5, 0.95, etc. If not specified, simply calcualte mean values. Used for spatial plots only.
```
rschwant, line 103: @jianheRCM can we change the location for this percentile_opt. It fits better in the plots section under data_proc. For example, all the other plotting/data analysis type metrics that are not variable specific are stored here instead. This would make it nice because if someone wanted to plot mean, 95th, and 99th percentile spatial bias plots they could just do this with one submit script and have three plotting groups with the different options. Does this make sense?

rschwant:
This looks good Jian. I just added one comment. For the NaNs, I see that the values are NaNed during the regulatory metric calculation so good point. I'm still thinking we need a warning to ensure people are setting rem_obs_nan = True appropriately. But this is probably a general comment beyond this regulatory application. I will think about a general warning for this and add it in later.

examples/yaml/control_rrfs_cmaq_airnow_reg.yaml
jianheRCM: Good point. Will put under data_proc.

jianheRCM:
> This looks good Jian. I just added one comment. For the NaNs, I see that the values are NaNed during the regulatory metric calculation so good point. I'm still thinking we need a warning to ensure people are setting rem_obs_nan = True appropriately. But this is probably a general comment beyond this regulatory application. I will think about a general warning for this and add it in later.

Becky, I have included the warning message. A few updates in yaml file and driver are also included.

jianheRCM:
@rschwant @zmoon I have included several updates in the yaml file and code. Could you please review them and let me know what you think? thanks!

zmoon (commented):
It's looking good.

Some general comments:

* In the plotting functions, could eliminate the `regulatory` option and just compute it from `df_reg` (i.e., `regulator = df_reg is not None`), the assumption being that if you pass `df_reg` you want to plot the regulatory stuff.

* Seems like a lot of extra prints of objects and such added in the driver

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +909,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+        df_counto = df[df[column_o]> 35.].groupby(['siteid'],as_index=False)[column_o].count()
+        df_countm = df[df[column_m]> 35.].groupby(['siteid'],as_index=False)[column_m].count()
+        ylabel2 = 'PM2.5'
+
+    # combine dataframes
+    df_combine = df_counto.set_index(['siteid']).join(df_countm.set_index(['siteid']),on=(['siteid']),how='outer').reset_index()
+    df_combine[column_o]=df_combine[column_o].fillna(0)
+    df_combine[column_m]=df_combine[column_m].fillna(0)
+
+    df_reg = df_mean.reset_index(drop=True).merge(df_combine.reset_index(drop=True),on=['siteid']).rename(index=str,columns={column_o+'_y':column_o+'_day',column_m+'_y':column_m+'_day'})
```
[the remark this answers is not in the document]
zmoon: from the merge I guess. Could use the `suffixes` kwarg to control that so don't have to rename

melodies_monet/plots/surfplots.py
```diff
@@ -39,8 +40,15 @@ def make_24hr_regulatory(df, col=None):
...
         
     """
-    return calc_24hr_ave(df, col)
+    #return calc_24hr_ave(df, col)
+    return calc_24hr_ave_v1(df, col)
 
+def calc_24hr_ave_v1(df, col=None):
+    df.index = df.time_local
+    # select sites with nobs >=12
+    df_24hr_ave = (df.groupby("siteid")[col].resample("D").sum(min_count=12)/df.groupby("siteid")[col].resample("D").count()).reset_index().dropna()
```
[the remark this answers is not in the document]
jianheRCM: I only found the requirement of 75% completeness for 8-hr ozone calculations. It could also be applied to 24-hr PM2.5 calculation as well. I will update this for PM2.5.

melodies_monet/plots/surfplots.py
```diff
@@ -138,8 +155,67 @@ def map_projection(f):
...
 
-def make_spatial_bias(df, column_o=None, label_o=None, column_m=None, 
-                      label_m=None, ylabel = None, vdiff=None,
+def get_utcoffset(lat,lon):
+    """get UTC offset in hour based on a point (lat/lon)
+
+    Parameters
+    ----------
+    lat : latitude of a single point
+    lon : longitude of a single point
```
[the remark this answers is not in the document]
jianheRCM: Should be 
Latitude (deg; -90. to 90.)

melodies_monet/plots/surfplots.py
```diff
@@ -138,8 +155,67 @@ def map_projection(f):
         raise NotImplementedError('Projection not defined for new model. Please add to surfplots.py')
     return proj
 
-def make_spatial_bias(df, column_o=None, label_o=None, column_m=None, 
-                      label_m=None, ylabel = None, vdiff=None,
+def get_utcoffset(lat,lon):
```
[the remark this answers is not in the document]
jianheRCM: Yes. make_spatial_bias is used to make spatial bias plots. See in the driver.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
zmoon: Yeah sorry, the -180 to 180 part was supposed to go in longitude, that's the important one to qualify

melodies_monet/plots/surfplots.py
```diff
@@ -138,8 +155,67 @@ def map_projection(f):
...
+    import datetime, pytz
+    from datetime import datetimee, timezone
+    import pytz
+    from timezonefinder import TimezoneFinder
+
+    tf = TimezoneFinder()
+
+    lon2 = lon.values.tolist()
+    lat2 = lat.values.tolist()
+    timezone_str = tf.timezone_at(lng=lon2, lat=lat2)
```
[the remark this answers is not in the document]
jianheRCM: Correct. I have updated the code.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
zmoon: I was pointing to the `get_utfcoffset` here.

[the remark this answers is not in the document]
jianheRCM: The function here will be used to make spatial overlay plots for MDA8O3 and 24hrPM2.5, which is not included at this point. The spatial plot for modeled MDA8 O3 requires to use local hours, which will use this function to convert UTC to local time for each model grids. This function really takes lot of time for the calculations so I have not included the spatial overlay plots for MDA8O3 and 24hrPM2.5. Becky mentioned a simplified way to do such conversion, which could be included later.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
zmoon: hmm well I tried it and it raised error saying out of bounds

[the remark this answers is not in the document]
jianheRCM: I think you are correct. I checked the user guide and it says
 lng–longitude of the point in degree (-180.0 to 180.0)
Thanks for correcting this. I will include in the code.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
zmoon: > This function really takes lot of time for the calculations so I have not included the spatial overlay plots for MDA8O3 and 24hrPM2.5. Becky mentioned a simplified way to do such conversion, which could be included later.

Right, we can do that later. I was thinking to include capabilities to do this with different methods (including the one Becky mentioned and the one you have) in monetio or monet,

melodies_monet/plots/surfplots.py
```diff
@@ -769,3 +909,145 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
+ 
+    elif column_o == 'PM2.5_reg':
+        df_mean=df.groupby(['siteid'],as_index=False).mean()
+        # get the exceedance days for each site
+        df_counto = df[df[column_o]> 35.].groupby(['siteid'],as_index=False)[column_o].count()
+        df_countm = df[df[column_m]> 35.].groupby(['siteid'],as_index=False)[column_m].count()
+        ylabel2 = 'PM2.5'
+
+    # combine dataframes
+    df_combine = df_counto.set_index(['siteid']).join(df_countm.set_index(['siteid']),on=(['siteid']),how='outer').reset_index()
```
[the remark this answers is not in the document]
jianheRCM: Somehow this change does not work. Below is the error message:
ValueError: You are trying to merge on object and int64 columns. If you wish to proceed you should use pd.concat

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
rschwant: I'm fine with leaving it here for now and depending on future development moving it to monetio or monet.

rschwant (commented):
Just have a couple minor thoughts now and I think this is near ready. What do you think about leaving the scaling for both ppb and OC/OM in the yaml file. I feel like people can just multiply by 1000 as needed to do ppb conversions from ppm and similar to OC/OM in the yaml file itself rather than adding options for this?

melodies_monet/driver.py
```diff
@@ -211,6 +213,11 @@ def mask_and_scale(self):
                             self.obj[v].data += scale
                         elif d['unit_scale_method'] == '-':
                             self.obj[v].data += -1 * scale
+                    if 'obs_unit' in d:
```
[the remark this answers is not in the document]
jianheRCM: In AQS data, O3 and other gases are in ppm. The reader scripts convert model ppm to ppb but not the obsdata.  obs_unit is not for all the tracers. User can simply include or not in the yaml file.

melodies_monet/driver.py
```diff
@@ -389,6 +396,10 @@ def mask_and_scale(self):
                             self.obj[v].data += scale
                         elif d['unit_scale_method'] == '-':
                             self.obj[v].data += -1 * scale
+                    if 'om2oc_factor' in d:
```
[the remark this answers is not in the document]
jianheRCM: I do not want to mess up with the "unit_scale" as it is also used for the obsdata conversion. I would rather convert model data to match with obsdata. Users can define OM_to_OC ratio in yaml file. If there is no need for the conversion, users can simply set the ratio to be 1 or do not include it in yaml file.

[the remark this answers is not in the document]
rschwant: Sounds good, I don't mind including extra options if you think users will prefer it over using the general scaling approach, which just to be sure you know you can do for either the observations or the model variables although we only show an example for how to apply the scaling for the observations. Let's just be sure to thoroughly explain these extra namelist options in the docs.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
rschwant: @jianheRCM It looks like this is the last item that needs to be addressed and then I am happy to approve this and we can merge to develop.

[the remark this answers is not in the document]
jianheRCM: Somehow we need to use set_index for both df_counto and df_countm to make it work. So i will leave it as it is for now.

rschwant (approved):
Looks great! Thanks for incorporating all of Zach and my comments.

zmoon:
Thanks @jianheRCM for being patient with all of my comments!

jianheRCM:
@zmoon @rschwant I just found an issue in surfplots.py when use percentile option. The initial purpose of this percentile option is to do percentile calculation for O3 only. This option percentile_opt is currently set in the yaml file under "data_proc" as suggested by Becky. However, it turned out this would lead to the percentile calculation for all the species.  So I have to fix surfplots.py to make sure we only do percentile calculation for O3.  However, I would still suggest put percentile_opt under obs/variables for individual species if we want to do the percentile calculation (or not), which can have more flexibility to choose different percentile for different species. Please let me know what you think. Thanks,

jianheRCM:
> @jianheRCM If you only want to apply the percentile option to MDA8 ozone I see why you would set it up this way. Is there a reason you do not want to set up an option where users could say plot the 95th percentile of PM2.5 or NO2 as well? With the old approach you could have both mean plots and 95th percentile plots in the same call by having two plotting groups in the yaml file. But with your new approach you would have to submit multiple yaml file scripts. For now, I don't mind limiting this feature to MDA8 ozone, but it would be good to have a general option eventually too.

@rschwant My intension is to include percentile options for other species as well. But I prefer to including different percentile options. For example, some users may want to calculate 95th percentile of O3 and 50th percentile of PM2.5 at the same time (in the same yaml file). In this case, we have to include the options under obs/variables  instead of data_proc. I have updated the code and yaml file to include such feathers.

jianheRCM:
> > @jianheRCM If you only want to apply the percentile option to MDA8 ozone I see why you would set it up this way. Is there a reason you do not want to set up an option where users could say plot the 95th percentile of PM2.5 or NO2 as well? With the old approach you could have both mean plots and 95th percentile plots in the same call by having two plotting groups in the yaml file. But with your new approach you would have to submit multiple yaml file scripts. For now, I don't mind limiting this feature to MDA8 ozone, but it would be good to have a general option eventually too.
> 
> @rschwant My intension is to include percentile options for other species as well. But I prefer to including different percentile options. For example, some users may want to calculate 95th percentile of O3 and 50th percentile of PM2.5 at the same time (in the same yaml file). In this case, we have to include the options under obs/variables instead of data_proc. I have updated the code and yaml file to include such feathers.

@rschwant we do not need to have different plotting groups for different percentiles.

examples/yaml/control_rrfs_cmaq_airnow_norm.yaml
```diff
@@ -0,0 +1,247 @@
...
+        vmin_plot: 15.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 55.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 20.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 21 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: False #Opt compute regulatory functions
+        ylabel_reg_plot: 'MDA8 O3 (ppbv)'
+        vmin_reg_plot: 30.0 #Opt
+        vmax_reg_plot: 70.0 #Opt
+        vdiff_reg_plot: 15.0 #Opt
+        #percentile_opt: 95 #Opt, 5, 50, 95, etc. If not specified, simply calcualte mean values. Currently used for "spatial_bias".
```
rschwant, line 107:
```suggestion
        #percentile_opt: 95 #Opt, 5, 50, 95, etc. If not specified, simply calculate mean values. Currently used for "spatial_bias".
```

examples/yaml/control_rrfs_cmaq_airnow_norm.yaml
```diff
@@ -0,0 +1,247 @@
...
+        vmin_plot: 0.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 22.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 15.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 23 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: False #Opt compute regulatory functions
+        ylabel_reg_plot: 'PM2.5_24hr (ug/m3)'
+        vmin_reg_plot: 0.0 #Opt
+        vmax_reg_plot: 22.0 #Opt
+        vdiff_reg_plot: 5.0 #Opt
+        #percentile_opt: 50 #Opt, 5, 50, 95, etc. If not specified, simply calcualte mean values. Currently used for "spatial_bias".
```
rschwant, line 92:
```suggestion
        #percentile_opt: 50 #Opt, 5, 50, 95, etc. If not specified, simply calculate mean values. Currently used for "spatial_bias".
```

examples/yaml/control_rrfs_cmaq_airnow_reg.yaml
```diff
@@ -0,0 +1,247 @@
...
+        vmin_plot: 0.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 22.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 15.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 23 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: True #Opt compute regulatory functions
+        ylabel_reg_plot: 'PM2.5_24hr (ug/m3)'
+        vmin_reg_plot: 0.0 #Opt
+        vmax_reg_plot: 22.0 #Opt
+        vdiff_reg_plot: 5.0 #Opt
+        #percentile_opt: 50 #Opt, for regulatory calculation of 24hrPM2.5, we simply calcualte mean values, no need to set percentile_opt
```
rschwant, line 92:
```suggestion
        #percentile_opt: 50 #Opt, for regulatory calculation of 24hrPM2.5, we simply calculate mean values, no need to set percentile_opt
```

examples/yaml/control_rrfs_cmaq_airnow_reg.yaml
```diff
@@ -0,0 +1,247 @@
...
+        vmin_plot: 15.0 #Opt Min for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vmax_plot: 55.0 #Opt Max for y-axis during plotting. To apply to a plot, change restrict_yaxis = True.
+        vdiff_plot: 20.0 #Opt +/- range to use in bias plots. To apply to a plot, change restrict_yaxis = True.
+        nlevels_plot: 21 #Opt number of levels used in colorbar for contourf plot.
+        regulatory: True #Opt compute regulatory functions
+        ylabel_reg_plot: 'MDA8 O3 (ppbv)'
+        vmin_reg_plot: 30.0 #Opt
+        vmax_reg_plot: 70.0 #Opt
+        vdiff_reg_plot: 15.0 #Opt
+        percentile_opt: 95 #Opt, 5, 50, 95, etc. If not specified, simply calcualte mean values. Currently used for "spatial_bias".
```
rschwant, line 107:
```suggestion
        percentile_opt: 95 #Opt, 5, 50, 95, etc. If not specified, simply calculate mean values. Currently used for "spatial_bias".
```

rschwant:
@jianheRCM Just updated 4 spellings of calculate. You or I can commit these and then I think this is ready to merge to develop. Do you agree?

examples/yaml/control_rrfs_cmaq_airnow_reg.yaml
jianheRCM: I just corrected. Thanks,

jianheRCM: I just corrected. Thanks,

PR #85, opened 2022-05-20
Dev note on install page

zmoon:
This adds a link to the developer's guide doc to the main install docs page.
@bbakernoaa suggested doing this at the last meeting.

PR #86, opened 2022-05-23
Merge develop

zmoon:
Summary:
* updated Setuptools config and metadata, in order to support installing `melodies_monet` as Python package
* initial CLI (#80)
* lazy top-level imports (part of #80)
* simplified install docs (#81, #85)


melodies_monet/__init__.py
```diff
@@ -1,12 +1,38 @@
 """
 MELODIES MONET
 """
-from . import driver, plots, stats, util, tutorial
+import sys
 
-__all__ = (
+__version__ = "0.1.dev"
```
zmoon, line 6: For main we should probably remove the dev. How about 0.1.0? @rschwant

zmoon: Keeping dev to indicate development release/version is fine too

rschwant: Let's keep dev for development for now. Once we have an official release to coordinate with publication of a paper we can take this off.

rschwant (approved):
This looks good Zach.

melodies_monet/__init__.py
zmoon: Ok, sounds good.

rschwant:
@dwfncar David, do you have time to review this? I want to push this to main before we pull the regulatory metrics into develop. Early next week works too. Thanks!

rschwant:
@rrbuchholz Hey Rebecca, do you have time to be a reviewer for these updates next week? I think you were involved in some of their development too and so are somewhat familiar. A lot of the updates pertain to the user guide. Thanks!

rschwant:
@zmoon I just realized that to make things consistent, we need to upload the examples here: 
examples/jupyter_notebooks/*
examples/submit_jobs/run_melodies_monet.py

to no longer include the line, which is no outdated:
import sys; sys.path.append("../../")

It looks like this is all updated well in the docs section for those examples. Can you update this? Thanks!

zmoon:
Thanks @rschwant , I just removed those lines

zmoon:
@rschwant changing that to develop is the right play right now. I doubt it is related to MM, I would suspect a cartopy or Hera change. I saw something in one of the UFS repos recently, I'll see if I can find that.

zmoon:
@rschwant I did a bit of investigation. It seems like `/scratch1/NCEPDEV/rstprod/nexus_emissions/cartopy_shapefiles` is perhaps no longer a good source for the shape files. When I ran the script all I got was this:

image

i.e., just 110-m land. So my guess is that any other links you had are broken now, so cartopy tries to download. I don't have permission to go inside `/scratch1/NCEPDEV/rstprod` so I can't investigate in there further.

rschwant:
@SarahLu-NOAA The back and forth with cartopy on Hera has been resolved, so this code is ready for review. Do you mind reviewing the code this week sometime? Thanks!

SarahLu-NOAA:
@rschwant 
Sure - will give it a try.

zmoon:
I've noticed some files not having the license header thing. I should put them in, and maybe put in a basic check in CI to check that at least all Python files have it.

rschwant:
Yes, that sounds good Zach.

SarahLu-NOAA:
@rschwant 
I'm reviewing the code/doc.  Just wonder whether MELODIES-MONET has standard regression test package?  If so, do I have to run the regression test as part of code review?

SarahLu-NOAA (approved):
I forgot to mark the files reviewed.  
The doc/code looks good.   I particularly like moving machine-specific install to the appendix.

rschwant:
Great thanks Sarah. Are we ready to merge this into main then @zmoon?

zmoon:
@rschwant link check failed on merge because <https://www2.acom.ucar.edu/workshop/musica-tutorial-2021> has been removed it seems

zmoon:
Thanks. A redirect would've been handy..

PR #87, opened 2022-06-23
Merge develop

zmoon:
Summary:

* [x] Update (fix) the MUSICA tutorial URL [^b]
* [x] CI updates (use provision-with-micromamba to make it faster [^a])

[^b]: changed sometime between #86's last commit and the merging of #86 (note link check failure for 10aca92)
[^a]: the docs build check was taking as long as 7 minutes before, now as short as 30 seconds... 🥳


rschwant (approved):
Looks good.

rschwant:
Hi @jianheRCM, do you have time to review these updates today for this pull request? There are only two very small changes. And once this moves to main, we can move your regulatory options to develop. Thanks!

jianheRCM (approved):

rschwant:
@zmoon I found one more problem that I think we need to fix, but wanted to make sure you agree. In this jupyter notebook I think we need to point to the melodies_monet.util folder
https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/process_obs/reformat_airnow.ipynb
So change
from util import write_util
to
from melodies_monet.util import write_util

I think all of these obs examples would need to be updated. Do you agree? I cannot remember now why we had changed this, but it may be just an artifact of all the reshuffling we have done recently and it used to work when these files were in a different location. 

The last line of all these files also needs to be removed, since we are no longer putting these files in a data folder.
ls -lh data/* 

I'll submit a pull request to develop and mark you as reviewer by Monday morning, but then let's get this into the main branch in this pull request too. Then we can push all of this to main before the meeting.

zmoon:
Hi @rschwant yes the imports should be updated. Those notebooks weren't updated when MM was made a Python package.

Your plan sounds good. We could also after that think about including them in the docs if you think that would be valuable. And/or in CI checks.

PR #88, opened 2022-06-27
Update reader scripts so that reads util from MELODIES-MONET correctly.

rschwant:
This updates the jupyter notebook examples to create the observations data to correctly read in util from MELODIES-MONET. Note Zach the AirNow reader is working well, but the AERONET and IMPROVE readers seem not to work properly anymore potentially due to changes in MONETIO and/or the observation data format. I'm adding a note to fix this in the Checklist, but I want to get this format change fixed for AirNow ASAP as others are having trouble with the reader.


zmoon:
> I feel like it's easier to keep track of the code if the outputs are not included

Yes, this is generally the way to go, for smaller and more informative diffs, unless the notebook takes a long time to run or special resources. Thanks.

zmoon (approved):

rschwant:
For notes, the last change was just to update an outdated directory name and save the output file in the current directory instead. And also included notes in header of file for how to load pytspack package.

PR #96, opened 2022-08-15
Make Dask progress bars optional

### LINKED ISSUE #95: Disable progress bars
Would be nice to provide an option to disable Dask progress bars (they make log files from a non-interactive run hard to read). [Seems like](https://stackoverflow.com/a/43121416) it is possible to unregister all that may have been registered (e.g. in monetio). Just need to implement a MM config option.

zmoon:
Resolves #95 

References:
* [Dask Diagnostics (local)](https://docs.dask.org/en/latest/diagnostics-local.html)
* [outdated SO answer about removing](https://stackoverflow.com/questions/43119440/how-to-unregister-all-progressbars/43121416#43121416)
* [PR describing new config methods](https://github.com/dask/dask/pull/3502)

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -475,6 +475,18 @@ class analysis:
             self.output_dir = self.control_dict['analysis']['output_dir']
         self.debug = self.control_dict['analysis']['debug']
 
+        # Enable Dask progress bars? (default: false)
+        enable_dask_progress_bars = self.control_dict["analysis"].get(
+            "enable_dask_progress_bars", False)
+        if enable_dask_progress_bars:
+            from dask.diagnostics import ProgressBar
+
+            ProgressBar().register()
+        else:
+            from dask.callbacks import Callback
+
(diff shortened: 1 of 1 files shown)


zmoon:
@rschwant I put this in `analysis.read_control()` method. Does that seem like a good place? Also are you ok with defaulting to false? Or do you want default true?

zmoon:
Don't understand the docs build failure. Same Sphinx version as when it was passing recently (4.5.0).

rschwant:
@zmoon Adding this to the analysis section seems like a good place to me. I also agree let's default to false and if people need to output these diagnostics they can set the value to true. I tested the commands in another jupyter notebook and they worked for me, so I can review this if needed too. Not sure either why the docs are failing now?

zmoon:
@rschwant thanks for testing the commands!

rschwant (approved):
Looks good!

zmoon:
Thanks @rschwant

PR #101, opened 2022-08-23
Add basic pairing for aircraft observations

colin-harkins:
- Added basic pairing functionality for aircraft observations. This was tested using firexaq aircraft observations but should function with others. 
- Added a .yaml file with a configuration for loading firexaq observations. 
- added functions to the observation and model classes (rename_vars) that will rename variables based on the .yaml. This is so that the pressure level in the model and observations can be set to a standard name ("pressure_obs" and "pressure_model"), to function with the vertical interpolation code. Additionally, nonstandard names for latitude and longitude can be renamed as needed.
- Plotting for timeseries, taylor and boxplot types all work correctly on model/aircraft paired data. Stats also works correctly on model/aircraft paired data.

### DIFF
--- a/examples/yaml/control_wrfchem_aircraft.yaml
+++ b/examples/yaml/control_wrfchem_aircraft.yaml
@@ -8,7 +8,6 @@ analysis:
   end_time: '2019-09-06-00:00:00' #UTC
   output_dir: /wrk/charkins/melodies_monet/aircraft/analysis #Opt if not specified plots stored in code directory.
   debug: True
-  resample: '600S' #6 min so works on Hera as a test. Can comment this if submitting a job.
 model:
   wrfchem_v4.2: # model label
     # files: /wrk/d2/rschwantes/wrf/firex_mech/qzhu/run_CONUS_fv19_BEIS_1.0xISO_RACM_v4.2.2_racm_berk_vcp_noI_phot_soa/0905/*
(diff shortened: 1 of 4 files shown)


melodies_monet/util/tools.py
```diff
@@ -271,3 +271,51 @@ def get_epa_region_df(df):
...
+    df_model = xr.merge(var_out_list).to_dataframe().reset_index()
+    df_model.pressure_model.fillna(df_model.pressure_obs,inplace=True)
+    df_model.drop(labels=['x','y','z','pressure_obs','time_obs'], axis=1, inplace=True)
+    df_model.rename(columns={'pressure_model':'pressure_obs'}, inplace=True)
+
+    final_df_model = merge_asof(df_obs, df_model, 
+                            by=['latitude', 'longitude', 'pressure_obs'], 
+                            on='time', direction='nearest')
+
+    return final_df_model
```
zmoon, line 321: missing newline at end of file

melodies_monet/driver.py
```diff
@@ -11,6 +11,7 @@
 import datetime
 
 # from util import write_ncf
+import melodies_monet
```
zmoon, line 14:
```suggestion
```
Suggest removing this and importing the function or module you need within the method that needs it. For example

```python
from .util.tools import vert_interp
```

melodies_monet/util/tools.py
colin-harkins: Added newline at end of file

melodies_monet/driver.py
colin-harkins: Made suggested changes

rschwant (commented):
This looks great Colin! I added just a couple suggestions. What do you think about these suggestions? Otherwise I think this is ready to merge into the aircraft branch.

examples/yaml/control_wrfchem_aircraft.yaml
```diff
@@ -0,0 +1,141 @@
...
+      linestyle: ':' 
+obs:
+  firexaq: # obs label
+    filename: '/wrk/d2/rschwantes/obs/firex-aq/R1/10s_merge/firexaq-mrg10-dc8_merge_20190905_R1.ict'
+    obs_type: aircraft
+    variables: #Opt 
+      'O3_CL_RYERSON':
+        unit_scale: 1 #Opt Scaling factor 
+        unit_scale_method: '*' #Opt Multiply = '*' , Add = '+', subtract = '-', divide = '/'
+        nan_value: -777777 # Opt Set this value to NaN
```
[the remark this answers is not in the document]
colin-harkins: I've added in a way to replace the LLOD in the latest commit. 

To be consistent with the use of nan_value in the yaml file, I used LLOD_value=-888888 (the value as it is in the original dataset) and LLOD_setvalue=0.0 (the value to replace -888888 with). If there's a better name than LLOD_setvalue that you can think of I can change it. I figured having LLOD_value correspond to the value after replacing and nan_value correspond to the value prior to replacing would be confusing.

I also set this up so that LLOD_setvalue replaces LLOD_value **after** any unit conversion occurs. This is also noted in a comment in the control file.

melodies_monet/driver.py
```diff
@@ -474,6 +510,8 @@ def read_control(self, control=None):
         if 'output_dir' in self.control_dict['analysis'].keys():
             self.output_dir = self.control_dict['analysis']['output_dir']
         self.debug = self.control_dict['analysis']['debug']
+        if 'resample' in self.control_dict['analysis'].keys():
```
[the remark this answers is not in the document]
colin-harkins: In the latest commit I moved the resampling to the obs class. In the control file, the resample value is now set under the type of obs. This would also allow for setting different resample intervals for different obs types if someone ever had a need for that.

PR #104, opened 2022-09-01
Added ability to save out and read in attributes of the analysis class (paired, models, obs).

colin-harkins:
- Added functions to save out attributes of the analysis class (paired, models, obs) in both pickle and netcdf format. Netdf files are saved such that groups within the attribute (i.e. a single model/obs pair) are saved as a group within the netcdf file. Further groups are added to the netcdf as necessary. The class attributes are saved to a json formatted string in the netcdf global attributes for each group.
- Added functions to read in attributes of the analysis class (paired, models, obs) that have been saved in both pickle and netcdf format. For netcdf files, the class attributes are repopulated from a json formatted string in the netcdf global attributes for each group. The appropriate analysis attribute is then set with the data (paired, model, obs)
- If multiple files are specified for read in, the read_saved_data function will automatically merge the datasets before setting the analysis attribute specified. 

In progress:
- need to test that reading in models and obs attributes functions correctly.
- need to test that reading in multiple netcdf files functions correctly.
- Netcdf files do not work correctly when using util.write_util.compress_variable so need to debug this

### DIFF
--- a/melodies_monet/util/write_util.py
+++ b/melodies_monet/util/write_util.py
@@ -22,8 +22,13 @@ def write_grouped_ncf(obj, output_name, title=''):
     """
     import pandas as pd
     import json
+    import os.path
     
     print('Writing:', output_name)
+    
+    if os.path.exists(output_name):
+        os.remove(output_name)
+    
     for group in obj.keys():
         print('Writing group:', group)
         dset=obj[group].obj
--- a/melodies_monet/util/write_util.py
+++ b/melodies_monet/util/write_util.py
@@ -4,6 +4,44 @@
 import numpy as np
 from pandas.api.types import is_float_dtype
 
+def write_grouped_ncf(obj, output_name, title=''):
+    """Function to write netcdf4 files with some compression for floats from an attribute of the
+    analysis class (models, obs, paired).
+
+    Parameters
+    ----------
+    obj : dict
+        Dict containing attribute of the driver analysis class (model, observation or paired).
+    output_name : str
+        Description of parameter `output_name`.
+
+    Returns
+    -------
+    None
+
+    """
+    import pandas as pd
+    import json
+    
+    print('Writing:', output_name)
+    for group in obj.keys():
+        print('Writing group:', group)
+        dset=obj[group].obj
+        comp = dict(zlib=True, complevel=7)
+        encoding = {}
+        for i in dset.data_vars.keys():
+            # if is_float_dtype(dset[i]):  # (dset[i].dtype != 'object') & (i != 'time') & (i != 'time_local') :
+            #     print("Compressing: {}, original_dtype: {}".format(i, dset[i].dtype))
+            #     dset[i] = compress_variable(dset[i])
+            encoding[i] = comp
+        dset.attrs['title'] = title
+        dset.attrs['format'] = 'NetCDF-4'
+        dset.attrs['date_created'] = pd.to_datetime('today').strftime('%Y-%m-%d')
+        dict_json = obj[group].__dict__.copy()
+        dict_json.pop('obj')
+        dset.attrs['dict_json'] = json.dumps(dict_json, indent = 4) 
+        # dset.to_netcdf(output_name, encoding=encoding,group=group,mode='a')
+        dset.to_netcdf(output_name, group=group, mode='a')
 
 def write_ncf(dset, output_name, title=''):
     """Function to write netcdf4 files with some compression for floats
@@ -132,3 +170,24 @@ def compress_variable(da):
     da.attrs['_FillValue'] = -1
     da.attrs['missing_value'] = -1
     return da
+
+def write_pkl(obj, output_name):
+    """Function to write a pickle file from an attribute of the analysis class (models, obs, paired)
+
+    Parameters
+    ----------
+    obj : type
+        Description of parameter `obj`.
+    output_name : str
+        Description of parameter `output_name`.
+        
+    Returns
+    -------
+    None
+
+    """
+    from joblib import dump
+    
+    print('Writing:', output_name)
+    with open(output_name, 'wb') as outp:
+        dump(obj, outp)
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -647,11 +647,9 @@ class analysis:
         if len(files)==1:
             if method=='pkl':
                 setattr(self, attr, read_pkl(files[0]))
-                 # self.paired = read_pkl(filename[0])
             elif method=='netcdf':
                 group_ds = read_grouped_ncf(files[0],xr_kws)
-                setattr(self, attr,  xarray_to_class(class_type=class_names[attr],group_ds=group_ds))
-                # self.paired = populate_class(class_type='pair',group_ds=group_ds)          
+                setattr(self, attr,  xarray_to_class(class_type=class_names[attr],group_ds=group_ds))        
         
         # If only >1 file then merge by group prior to setting
         elif len(files)>1:
@@ -673,7 +671,6 @@ class analysis:
                         attr_out[group].obj = xr.merge([attr_out[group].obj,attr_append[group].obj])
             
             setattr(self, attr,  attr_out)
-            # self.paired=paired_out
         
     
     ### TODO: Create the plotting driver (most complicated one)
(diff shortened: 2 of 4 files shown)


zmoon:
Some possible next steps:
* function in driver to save things to the configured output directory
* config settings in the YAML to control what should be saved and using what method
  - maybe under the `analysis` section
* maybe another config setting to decide whether to try and use potential saved datasets when loading or ignore them

melodies_monet/util/write_util.py
```diff
@@ -4,6 +4,49 @@
 import numpy as np
 from pandas.api.types import is_float_dtype
 
+def write_grouped_ncf(obj, output_name, title=''):
```
zmoon, line 7: Seems like this is potentially writing more than one nc file but using the same name for all. Might be simpler just to default to using the key for the name (at least the stem, e.g. `airnow`, `airnow_wrfchem`). Could offer an option to apply a suffix to all of the stems to allow for some control.

zmoon: Update: @colin-harkins has informed me that this writes a single file using netCDF4 [groups](https://unidata.github.io/netcdf4-python/#groups-in-a-netcdf-file).

colin-harkins:
@rschwant @zmoon Would one of you take a look at this when you have the chance and let me know what you think?

I changed the netcdf saving so that it does not use netCDF4 groups and instead saves each pair as a separate file. Additionally, I added in the ability to save data based on options in the control .yaml file (by running an.save_analysis()). I also added in the avility to read previously saved data based on options in the control .yaml file (by running an.read_analysis()).

colin-harkins:
@rschwant @zmoon I think this is ready for review. 

In the last 4 commits I've done the following: 

- Added example jupyter notebooks to both the examples/jupyter_notebooks and docs/examples folders
- Updated the docs to include the additional YAML options I've created for saving and reading files
- Added a test to avoid mistakes when reading in multiple netCDF files to the same pair/group. The code now raises an error if the pair name is not the same between netCDF files, instead of trying to merge the potentially conflicting xarray objects.

examples/yaml/control_saveandread.yaml
```diff
@@ -0,0 +1,117 @@
...
+          method: 'netcdf' # 'netcdf' or 'pkl'
+          prefix: '0905' # use only with method=netcdf
+          # output_name: '0905.pkl' # use only with method=pkl
+          data: 'all'   # 'all' to save out all pairs or ['pair1','pair2',...] to save out specific pairs. With method='pkl' this is ignored and always saves all.
+      # models:
+      # obs:
+  read:
+      paired:
+          method: 'netcdf' # 'netcdf' or 'pkl'
+          filenames: {'airnow_wrfchem_v4.2':['0905_airnow_wrfchem_v4.2.nc4']} # example for netcdf method. Uses dict of form {group1: str or iterable of filenames, group2:...}
```
rschwant, line 22: I guess my comment above is really does it default to using the output_dir to read in the output and if so, why does the ReadTheDocs example look different and include the entire directory name?

rschwant:
I thought of a good solution to that above. Add optional folders for the following:
output_dir_read
output_dir_write
Then if you don't provide anything for these they default to output_dir. This gives more flexibility, but is also very clean if people just want to store everything in the same place in their output_dir because they can just leave these out of their yaml file.
What do you think Colin? We can discuss it next next Monday.

examples/yaml/control_saveandread.yaml
colin-harkins: So currently it will always save to output_dir but reading I left such that it can be read from anywhere. In this example it is reading from the code directory and in the prior comment it is reading from the output_dir that was set in the yaml file, but with a relative address.

I like your solution of having optional 'output_dir_read' and 'output_dir_write' options and then defaulting to output_dir if they are not specified.

My only concern with defaulting to output_dir if output_dir_read is not specified that it limits the user to reading files from a single directory. I could add the ability to do something like `output_dir_read: None`, which would ignore the default directory when reading and instead read directly from the path specified in filenames: .

docs/examples/control_wrfchem_saveandread.yaml
```diff
@@ -0,0 +1,232 @@
...
+          prefix: '0905' # use only with method=netcdf
+          # output_name: '0905.pkl' # use only with method=pkl
+          data: 'all'   # 'all' to save out all pairs or ['pair1','pair2',...] to save out specific pairs. With method='pkl' this is ignored and always saves all.
+      # models:
+      # obs:
+  read:
+      paired:
+          method: 'netcdf' # 'netcdf' or 'pkl'
+          filenames: {'airnow_RACM_ESRL':['./output/save_and_read/0905_airnow_RACM_ESRL.nc4'],'airnow_RACM_ESRL_VCP':['./output/save_and_read/0905_airnow_RACM_ESRL_VCP.nc4']} # example for netcdf method. Uses dict of form {group1: str or iterable of filenames, group2:...}
+          # filenames: {'airnow_wrfchem_v4.2':['0904_airnow_wrfchem_v4.2.nc4','0905_airnow_wrfchem_v4.2.nc4']} # example for netcdf method. Uses dict of form {group1: str or iterable of filenames, group2:...}
```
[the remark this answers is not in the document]
colin-harkins: Let me test whether in yaml you can split a dict across multiple lines. If you can then I'll show how to do that here. Otherwise I'll figure out a way to have the filenames split across multiple lines because that is important.

I don't think it would be hard to implement wildcards. I'll just use glob.glob() to expand each element of the list with anything returned by glob.

docs/examples/control_wrfchem_saveandread.yaml
```diff
@@ -0,0 +1,232 @@
...
+#   - 'ty_scale' is needed for 'taylor' plots.
+#   - 'nlevels' or the number of levels used in the contour plot can also optionally be provided for spatial_overlay plot.
+#   - If set_axis = True and the proper limits are not provided in the obs section,
+#     a warning will print, and the plot will be created using the default limits.
+analysis:
+  start_time: "2019-09-05-06:00:00" # UTC
+  end_time: "2019-09-06-06:00:00" # UTC
+  output_dir: ./output/save_and_read  # relative to the program using this control file
+  # ^ optional; if not specified, plots are stored in code directory (??)
+  #   Currently, the directory must exist or plot saving will error and fail.
```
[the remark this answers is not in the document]
colin-harkins: This comment was just copied from control_wrfchem.yaml in the same directory, so I hadn't thought about it originally but sure, I can do this.

examples/yaml/control_saveandread.yaml
rschwant: I like this. Let's add this additional capability that if you can set output_dir_read = None, then you can just add the file paths separately for each filename. Let's just explain this in the yaml description in the docs since it's a bit complicated.

docs/examples/control_wrfchem_saveandread.yaml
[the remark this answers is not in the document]
rschwant: Great thanks for fixing this.

docs/examples/control_wrfchem_saveandread.yaml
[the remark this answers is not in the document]
colin-harkins: I checked and it is possible to split a dict over multiple lines in the .yaml file, so no issues there.

examples/yaml/control_saveandread.yaml
colin-harkins: As it turns out, you need to set it as `output_dir_read: null` in the yaml file so that python reads it as `output_dir_read = None` in the `control_dict`. I'll explain in the docs.

colin-harkins:
In the above commits I added:

- ability to use wildcards when reading in files
- optional yaml arguments controlling read/save directories. Defaults to output_dir when not specified.  Instead of using the name output_dir_write, I used the name output_dir_save so that it matches with the use of the name save under analysis in the control yaml. Let me know if you think output_dir_write would be better.
- Made output_dir non-optional in the control yaml

I still need to update the docs for these last changes as well as update the example notebooks and .yaml files. I should be able to do that quickly on Monday.

rschwant (approved):
This looks great! I like using save instead of write for consistency, so good idea.

PR #105, opened 2022-09-08
Site analysis scripts

mmarnold8:


rschwant:
This looks great Mackenzie! (@mmarnold8) I think a lot of users will find these scripts useful. I noticed that in this file daily.monet-eval_sites.sh there were a couple python scripts that are not uploaded that might also be useful. 

reformat_aeronet_rapchemtest.py
reformat_airnow_rapchemtest.py
Monet-analysis-example-plots-wrf-rapchemtest_aeronet.py
Monet-analysis-example-plots-wrf-rapchemtest.py

Would it be possible to upload these as well? There's even more scripts than I was expecting. Do you want to move these into a new folder within the scripts folder called: Forecast_Evaluation_Example (feel free to change this to what you feel is most appropriate) and move all these files there. That way everyone will know they belong together. Then do you mind adding a README that just briefly introduces all of these scripts and provides a link to their output on the RAP-chem page. Thanks!

zmoon:
I second the suggestion of @rschwant to group them together in a folder. And I think it would make more sense to put it under `examples` since it overall seems like a specific setup in current form, not a general tool. And lowercase the new folder name for consistency.

rschwant:
I'm good with that setup too. So Mackenzie (@mmarnold8), I change my mind now that I'm seeing that there are a large number of connected scripts you and Jordan have built. Let's put it in examples/forecast_evaluation instead. And if you can upload those additional .py files that would be great too! I'll update the user guide to point to this folder as I already have a link explaining the forecast evaluation you all have been doing as an example of MELODIES MONET in action. If you don't have time to add a README that is okay because we can eventually describe the scripts on the readthedocs page too.

mmarnold8:
That all sounds great! I can make those changes.

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
```diff
@@ -0,0 +1,62 @@
...
+import xarray as xr
+import monetio as mio
+from melodies_monet.util import write_util
+
+# Here is an example of how you would use/import a variable from the master script
+# ... python reformat_airnow.py $START_TIME $END_TIME ...
+import sys
+start_time_reformat=sys.argv[1]
+end_time_reformat=sys.argv[2] 
+print(sys.argv[1])
```
dwfncar, line 17: We could also use the argparse module for command line arguments.  It allows for flexible ordering and optional arguments with defaults.  Here is a typical example of its use:

    import argparse

    parser = argparse.ArgumentParser()
    parser.add_argument('--datadir', type=str,
        default=os.path.join(os.getenv('HOME'), 'Data', 'AirNow'),
        help='top-level data directory (default $HOME/Data/AirNow)')
    parser.add_argument('--debug', action='store_true',
        help='set logging level to debug')
    parser.add_argument('--start', type=str,
        default = datetime.now().strftime('%Y%m%d'),
        help='start date (yyyymmdd)')
    parser.add_argument('--end', type=str,
        default = datetime.now().strftime('%Y%m%d'),
        help='end date (yyyymmdd)')
    args = parser.parse_args()

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
```diff
@@ -0,0 +1,62 @@
...
+
+# add siteid back as a variable and create x as an array of integers
+tt['siteid'] = (('x'),tt.x.values)
+tt['x'] = range(len(tt.x))
+# expand dimensions so that it is (time,y,x)
+t = tt.expand_dims('y').set_coords(['siteid','latitude','longitude']).transpose('time','y','x')
+
+
+#wite out to filename set in cell 2
+write_util.write_ncf(t,'test5.nc')
```
dwfncar, line 62: Perhaps have the output file specified by the user in the command line arguments?
If using argparse
parser.add_argument('--outfile', type=str, help='output file')

default could be something like AirNow_starttime_endtime.nc.

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
```diff
@@ -0,0 +1,62 @@
+#!/usr/bin/env python
```
dwfncar, line 1: I suggest renaming the file to just reformat_airnow.py, as it is not specific to rapchem ...

zmoon:
Notes on the CI failure: https://github.com/noaa-oar-arl/monet/issues/115

rschwant:
During the last meeting we discussed that these codes would stay as is in the examples folder to document how RAP-chem and other research/operational models are being evaluated using MELODIES MONET at NOAA. Zach and David will work on generalizing these scripts to include within the command line setup in the main part of the code. @zmoon when you get a chance, can you check why the CI tests are failing? Thanks!

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
rschwant: Good point David, but given our discussion in the last meeting, let's keep all the RAP-chem specific files in this directory as is and you and Zach can work on creating generalized versions to include in the main code as a separate pull request.

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
rschwant: Thanks David. I like this idea. Let's do this for the more generalized versions of these codes.

examples/forecast_evaluation/reformat_airnow_rapchemtest.py
rschwant: Great idea. Let's add this in the more generalized version of this code.

rschwant (approved):
Looks great! Thanks @mmarnold8 and @jordanschnell for uploading these scripts!

PR #106, opened 2022-09-12
CLI updates - get AERONET and AirNow data

zmoon:
Adding (sub-)commands for downloading AERONET and AirNow.
* [x] AERONET download + format tool
* [x] Fix AERONET `daily` option with interp (raise error gracefully if monetio <= than current release)
* [x] AirNow download + format tool
* [x] Fix CLI doc if necessary
* [x] Change install doc reqs for CLI to `typer` instead of `click` and mention `rich` for color

#80 Added a command line interface (CLI) for running a MM control file, including typical steps that would be done if using the MM API. This PR adds some subcommands, allowing the MM CLI to do other things (download and format AERONET and AirNow data for use in MM). To get the previous behavior, you now have to use the `run` subcommand, i.e. `melodies-monet control.yml` ⟶ `melodies-monet run control.yml`, which makes more sense anyway.

Related to https://github.com/NOAA-CSL/MELODIES-MONET/issues/117, resolves partially https://github.com/NOAA-CSL/MELODIES-MONET/issues/118

### DIFF
--- a/melodies_monet/_cli.py
+++ b/melodies_monet/_cli.py
@@ -135,7 +135,7 @@ def get_aeronet(
     out_name: str = typer.Option(None, "-o",
         help=(
             "Output file name (or full/relative path). "
-            "By default the name is generated like 'AERONET_<product>_<start date>_<end date>.nc'"
+            "By default the name is generated like 'AERONET_<product>_<start-date>_<end-date>.nc'"
         )
     ),
     dst: Path = typer.Option(".", "-d", "--dst", help=(
--- a/melodies_monet/_cli.py
+++ b/melodies_monet/_cli.py
@@ -226,6 +226,8 @@ def get_aeronet(
     ]
 
     with _timer("Forming xarray Dataset"):
+        df = df.dropna(subset=["latitude", "longitude"])
+
         # Site-specific variables should only vary in x.
         # Here we take the first non-NaN value (should all be same).
         ds_site = (
@@ -238,7 +240,6 @@ def get_aeronet(
 
         ds = (
             df
-            .dropna(subset=["latitude", "longitude"])
             .set_index(["time", "siteid"])
             .to_xarray()
             .rename_dims(siteid="x")
--- a/melodies_monet/_cli.py
+++ b/melodies_monet/_cli.py
@@ -140,14 +140,114 @@ def get_aeronet(
     ),
     dst: Path = typer.Option(".", "-d", "--dst", help=(
             "Destination directory (to control output location "
-            "if using default out name)."
+            "if using default `out_name`)."
         )
     ),
     num_workers: int = typer.Option(1, "-n", "--num-workers", help="Number of download workers."),
 ):
     """Download AERONET data using monetio and reformat for MM usage."""
-    ...
+    import monetio as mio
+    import numpy as np
+    import pandas as pd
+    import xarray as xr
+
+    from .util.write_util import write_ncf
+
+    def expand_dims(ds, index=0, site_variable=None):
+        from numpy import unique
+
+        # First set a new index for the siteid
+        ds["x"] = index
+        ds = ds.expand_dims(["x"])
+        ds = ds.set_coords(["x"])
+
+        # Now reduce the site variables to single element variables
+        for sv in site_variable:
+            tmp = [unique(ds[sv])[0]]
+            ds[sv] = (("x",), tmp)
+
+        return ds
+
+    start_date = pd.Timestamp(start_date)
+    end_date = pd.Timestamp(end_date)
+    dates = pd.date_range(start_date, end_date, freq="D")
+
+    # Set destination and file name
+    fmt = r"%Y%m%d"
+    if out_name is None:
+        out_name = f"AERONET_L15_{start_date:{fmt}}_{end_date:{fmt}}.nc"
+    else:
+        p = Path(out_name)
+        if p.name == out_name:
+            # `out_name` is just the file name
+            out_name = p.name
+        else:
+            # `out_name` has path
+            if dst != Path("."):
+                print(f"warning: overriding `dst` setting {dst.as_posix()!r} with `out_name` {p.as_posix()!r}")
+            dst = p.parent
+            out_name = p.name
+
+    #standard_wavelengths = np.array([0.34, 0.44, 0.55, 0.66, 0.86, 1.63, 11.1]) * 1000.0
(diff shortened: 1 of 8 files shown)


zmoon:
Notes on the CI failure: https://github.com/noaa-oar-arl/monet/issues/115

zmoon:
Update: CI now passing with dev monet

melodies_monet/_cli.py
```diff
@@ -137,7 +137,7 @@ def get_aeronet(
     start_date: str = typer.Option(..., "-s", "--start-date", help="Start date."),
     end_date: str = typer.Option(..., "-e", "--end-date", help="End date."),
     daily: bool = typer.Option(False, help="Whether to retrieve the daily averaged data product."),
-    freq: str = typer.Option(None, "-f", "--freq", help="Frequency to resample to."),
+    freq: str = typer.Option("H", "-f", "--freq", help="Frequency to resample to."),
```
zmoon, line 140: with this change, there is no way to get the raw data product, but I think that is ok

zmoon:
Lot of overlap between the AERONET and AirNow ones. Maybe should try to DRY somewhat.

melodies_monet/_cli.py
```diff
@@ -126,6 +132,303 @@ def main(
             an.stats()
 
 
+@app.command()
+def get_aeronet(
```
zmoon, line 136: Should provide option to specify different product

rschwant:
@zmoon This looks great! I tested all the options and looked through the code. I have a couple suggestions/questions.

1) For run: I tested the melodies-monet run function on Hera as an easier way to submit a job rather than calling a separate python file and it worked well. I also really like the description you provide for each step. I like this option and if you have time, it would be great to provide an example submit script using the command line interface for users who want an even simpler method for submitting a job on Hera. The submit script I tested and worked well is here: /scratch2/BMC/rcm1/rhs/MONET/main/wrf-chem/qindan_test/submit_hera_cli_test.sh 

2) For get-airnow: I really like the nice help window. Can you add to the help window what format you expect the dates to be in for the observational readers. I assumed it was 2019-09-01 type format and it worked, but this might confuse some people. I also like that the units are now correctly represented in the netcdf file with your cli approach. Do you think it’s useful to keep the jupyter notebook and this method or once these cli versions are working well for all the observations, do you think we should just remove the jupyter notebook as an option? I'm trying to reduce the amount of code we need to maintain and I like your cli method better. We can keep both temporarily, but let's consider moving everything to cli in the future.

3) For get-aeronet: I also like the help window. Adding the date format here would also help. An example format for the wavelengths would also be useful as at first I was not exactly sure if I needed brackets, spaces, etc. and this would avoid confusion. Can you also add a link to the pytspack Readme here so that people know how to download it. I wonder if on this pytspack Readme it would be useful to specify what to do to download it on Hera and other linux machines at the top. I think that adding the two most likely ones to work first and then go into the details may be better like I did at the top of this jupyter notebook: https://github.com/NOAA-CSL/MELODIES-MONET/blob/develop/examples/process_obs/reformat_aeronet.ipynb There are just a lot of options in your readme and could be confusing to new users. 

4) For both AirNow and AERONET for your command line tools and the jupyter notebooks, I just realized you wrote that in the help guideline that the start and end date are closed on both sides, which is what I thought was the case too. But having looked at these files in more detail during my QC, it actually looks like you get the full first day and only the first hour or so of the last day. How do you make sure to get the full last day output with these methods? I think the output we have so far is basically truncating the last day. We could just have these examples go to the first day of the next month, but then we should be clear that the first day is closed and the last day is open, so that people choose the right dates for their time period. Do you want to change monetio to be closed on the last day or just recommend people choose a date range of 2019-09-01 to 2019-10-01 if they want the full month of September for example?

Thanks for doing all of this! This is a great addition especially for making the tool easier to use! Pending some discussion/updates on these points, I think this is ready to add into our main branch and I'll approve it.

zmoon:
Thanks very much @rschwant for these detailed comments!

Re: times, currently if you don't specify hour, 00 is assumed, which is why you don't get the full day if asking for hourly data (but you still get the day for daily data). But I agree if you don't specify hour you probably really want the whole day, so I can make that change. (Won't require monetio change, just which datetimes we pass to monetio.)

zmoon:
linkcheck failure was NCAR JupyterHub (currently down)

dwfncar:
_python setup.py install_

fails to find monetio

Installed /home/fillmore/miniconda3/envs/dev/lib/python3.9/site-packages/melodies_monet-0.1.dev0-py3.9.egg
Processing dependencies for melodies-monet==0.1.dev0
Searching for monetio
Reading https://pypi.org/simple/monetio/
Couldn't find index page for 'monetio' (maybe misspelled?)
Scanning index of all packages (this may take a while)
Reading https://pypi.org/simple/
No local packages or working download links found for monetio
error: Could not find suitable distribution for Requirement.parse('monetio')

zmoon:
> fails to find monetio

@dwfncar monetio isn't available on PyPI. Follow this sort of method instead: https://melodies-monet.readthedocs.io/en/latest/develop/developers_guide.html#setting-up-your-development-environment (note getting develop version of monet/monetio isn't necessarily required)

dwfncar:
Tested get-airnow.  All OK.

(dev) fillmore@Orion:~$ melodies-monet get-airnow --debug --start-date 20230131 --end-date 20230131 -o /home/fillmore/Data/AirNow/AirNow_20230131.nc
Please install s3fs if retrieving from the Amazon S3 Servers.  Otherwise continue with local data
Please install h5py to open files from the Amazon S3 servers.
Please install h5netcdf to open files from the Amazon S3 servers.
------------------
| MELODIES MONET |
------------------
Fetching data with monetio ...
Aggregating AIRNOW files...
Building AIRNOW URLs...
    Adding in Meta-data
Fetching data with monetio succeeded in 1.56 seconds
Forming xarray Dataset ...
Forming xarray Dataset succeeded in 0.0453 seconds
Writing netCDF file ...
Writing netCDF file succeeded in 0.247 seconds
(dev) fillmore@Orion:~$ 
(dev) fillmore@Orion:~$ 
(dev) fillmore@Orion:~$ ls ~/Data/AirNow/
AirNow_20230131.nc

zmoon:
@rschwant could you look at my recent commits and see if I addressed your comments sufficiently?

rschwant (approved):
Yes, this looks great. Thanks for making these updates.

PR #130, opened 2022-10-20
omps pairing and updates to plotting for xarray usage

mebruckner:
Updates RAQMS reader syntax to use verison of RAQMS reader discussed with Zach.
OMPS nadir mapper column ozone pairing
Draft changes to driver and satplots to keep satellite and model pairs in xarray format.


rrbuchholz:
Had a quick look - looks good to me.

PR #133, opened 2022-10-25
Adding MOPITT read via MONETIO

rrbuchholz:
Adding the MOPITT DataSet reader within MONETIO framework.


rrbuchholz:
Thanks Maggie. Hmmm I thought I had added a close command - thanks for catching that. You mean in the MONETIO fork code right:
https://github.com/NCAR/monetio/blob/develop_satellite/monetio/sat/_mopitt_l3_mm.py

mebruckner:
Yes in the MONETIO fork, around line 104 or earlier such as around line 89 when you finish reading things in from the file.

rrbuchholz:
Added a close commend to the hdf reading function. Will merge.

PR #143, opened 2022-11-16
Develop time chunks

dwfncar:
- In analysis.read_control, generate time_intervals from a config entry `control_dict['analysis']['time_interval']`, which should be a pandas freq string.
- Added optional argument `time_interval  = [Timestamp, Timestamp]` for start and end times to model and obs readers.
- Added method hook analysis.concat_pairs.

Changes should have no impact on current functionality of driver.
Intended for use in a top level script for a loop over time intervals:

```
from melodies_monet import driver

an = driver.analysis()
an.control = 'control_time_chunk.yaml'
an.read_control()

for time_interval in an.time_intervals:

    an.open_models(time_interval=time_interval)
    an.open_obs(time_interval=time_interval)
    an.pair_data()

an.concat_pairs()

an.plotting()
an.stats()
```


dwfncar:
Note individual readers that will support time chunking will need to implement the restriction to the time interval [start, end].

zmoon:
@dwfncar couldn't this be after the read step? We are expecting everything to be loaded as a xarray Dataset with a 'time' coordinate, so could just do the time selection after invoking the reader?

melodies_monet/driver.py
```diff
@@ -459,6 +470,14 @@ def read_control(self, control=None):
                 self.control_dict['analysis']['output_dir'])
         self.debug = self.control_dict['analysis']['debug']
 
+        # generate time intervals for time chunking
+        if 'time_interval' in self.control_dict['analysis'].keys():
+            time_stamps = pd.date_range(
```
colin-harkins, line 475: This isn't necessarily an issue, but start_time and end_time must be separated by an integer multiple of time_interval. If that isn't the case then whole time period will not be captured within self.time_intervals. 

As an example say start_time='2019-09-05-00:00:00' and end_time='2019-09-16-00:00:00' and time_interval='1D'. The final the value of self.time_intervals is [2019-09-15 00:00:00', '2019-09-16 00:00:00'], so the final 16 hours are missed. It may just be worth adding a note about this somewhere so that users are aware of that.

melodies_monet/driver.py
```diff
@@ -409,6 +419,7 @@ def __init__(self):
         """dict : Paired data, set by :meth:`pair_data`."""
         self.start_time = None
         self.end_time = None
+        self.time_intervals = None
```
colin-harkins, line 422: - self.time_intervals should be returned by __repr__(self) in analysis (around line 436).

melodies_monet/driver.py
zmoon: @colin-harkins did you mean end_time='2019-09-16-**16**:00:00'?

colin-harkins: @zmoon Yes, I did mean end_time='2019-09-16-16:00:00'. Thanks for catching that.

dwfncar: We could move the readers outside of the time interval loop, and

```
from melodies_monet import driver

an = driver.analysis()
an.control = 'control_time_chunk.yaml'
an.read_control()

an.open_models()
an.open_obs()

for time_interval in an.time_intervals:
    an.pair_data(time_interval=time_interval)

an.concat_pairs()

an.plotting()
an.stats()
```

dwfncar: But perhaps keep time_interval as an argument to the open_models and open_obs for now as well, to keep the flexibility in how to structure the top level time loop?

melodies_monet/driver.py
dwfncar: Agree with @colin-harkins, we do not want to truncate the complete analysis period.  In that example the fractional day should be included as the last time interval ... will address this ...

colin-harkins (approved):
Thanks for making these changes! This looks good to me.

PR #145, opened 2022-12-01
Clean up indentation in driver.py for plotting and statistics

colin-harkins:
In this commit, I cleaned up the indentation of the code originating from if statements on line 803, 1239 (in the commit). Originally the code was formatted like:

``` 
if pairdf.empty:  
    *print warning*  
else:  
    *proceed with plotting or stats execution*
```

It is now formatted like:
```
if pairdf.empty:
    *print warning*
    continue

*proceed with plotting or stats execution*
```

For some reason the diff shown below does not pick up that the lines originally under the else statement were just untabbed. It instead shows the diff as a large series of additions and deletions. I wasn't able to figure out a way to get the diff to pick up the untabbing so that the changes are more easily readable. Let me know if anyone has thoughts on this.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -802,206 +802,158 @@ class analysis:
                         # JianHe: do we need provide a warning if pairdf is empty (no valid obsdata) for specific subdomain?
                         if pairdf.empty or pairdf[obsvar].isnull().all():
                             print('Warning: no valid obs found for '+domain_name)
-                        else:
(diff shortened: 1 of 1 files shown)


rschwant (approved):
This looks great thanks for fixing this. I think the exact reason GitHub struggles to identify the changed tabs is why this is leading to confusing merge conflicts for the develop_satellite branch. I approve this can be merged into develop.

Issue #150, opened 2022-12-19
Restrict sites to include only those with enough data


rschwant:
RAP-chem team put in a restriction to only include sites with enough data points. I'd like to include this as an option in the yaml file too. The way RAP-chem team did it is here on line 133:

https://github.com/NOAA-CSL/MELODIES-MONET/blob/develop/examples/forecast_evaluation/site_analysis_airnow.py

Let's add something similar into the driver with the percentage of sites as a yaml file option.

rschwant:
With the pull request listed above by @colin-harkins this is complete and I am closing it.

PR #155, opened 2023-01-07
Restrict sites to include only those with enough data

colin-harkins:
Pull request to address issue #150.

This adds a yaml option to restrict observations based on the percentage of NaN values. The yaml option is formated as follows and lives under the data_proc option in the plotting and statistics groups:
`rem_obs_by_nan_pct: {'group_var': 'siteid','pct_cutoff': 25,'times':'hourly'}`

Originally this was intended to restrict by the percentage of data available for a site in airnow over the analysis time period. To make this more configurable I've made it so that the yaml option is configurable such that the user can remove by variables other that siteid in observation data. For example if the user wanted to remove any state where the NaN percentage is too great, the value for 'group_var' could be changed to 'state_name'.

In the yaml option 'pct_cutoff' is the maximum allowable percent of NaN values. The value for 'times' controls when the observations should be examined for NaN values. When 'times' is set to 'hourly', 'group_var' will only be examined for percentage NaN values at the top of the hour. When it is set to anything else it will look at all time values available. Airnow data seems to generally only be available on the hour but has NaN values between the hour so without restricting to on the hour airnow data has a much larger percentage NaN values.

### DIFF
--- a/docs/appendix/yaml.rst
+++ b/docs/appendix/yaml.rst
@@ -288,11 +288,11 @@ observation label is first and the model label is second
 
    * **rem_obs_by_nan_pct:** Specify as dictionary with keys 'group_var', 
      'pct_cutoff' and 'times'. If specified, removes all instances of 
-	 'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
-	 with airnow sites, setting 'group_var' to 'siteid' will remove all 
-	 sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
-	 only look at values at the beginning of each hour. Set 'times' to ''
-	 if all times should be used.
+     'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
+     with airnow sites, setting 'group_var' to 'siteid' will remove all 
+     sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
+     only look at values at the beginning of each hour. Set 'times' to ''
+     if all times should be used.
    * **rem_obs_nan:** If True, remove all points where model or obs variable is 
      NaN. If False, remove only points where model variable is NaN.
    * **set_axis:** If = True, use the axis constraints described in the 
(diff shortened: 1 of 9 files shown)


docs/appendix/yaml.rst
```diff
@@ -286,6 +286,13 @@ observation label is first and the model label is second
 
 **data_proc:** This section stores all of the data processing information.
 
+   * **rem_obs_by_nan_pct:** Specify as dictionary with keys 'group_var', 
+     'pct_cutoff' and 'times'. If specified, removes all instances of 
+	 'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
+	 with airnow sites, setting 'group_var' to 'siteid' will remove all 
+	 sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
+	 only look at values at the beginning of each hour. Set 'times' to ''
+	 if all times should be used.
```
zmoon, line 295: align this block with L edge of previous line to (hopefully) please the docs build CI

docs/appendix/yaml.rst
```diff
@@ -350,3 +357,13 @@ where domain_type is equal to domain_name.
...
+
+**data_proc:** This section stores all of the data processing information.
+
+   * **rem_obs_by_nan_pct:** Specify as dictionary with keys 'group_var', 
+     'pct_cutoff' and 'times'. If specified, removes all instances of 
+	 'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
+	 with airnow sites, setting 'group_var' to 'siteid' will remove all 
+	 sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
+	 only look at values at the beginning of each hour. Set 'times' to ''
+	 if all times should be used.
```
zmoon, line 369: this one too

docs/appendix/yaml.rst
```diff
@@ -286,6 +286,13 @@ observation label is first and the model label is second
 
 **data_proc:** This section stores all of the data processing information.
 
+   * **rem_obs_by_nan_pct:** Specify as dictionary with keys 'group_var', 
+     'pct_cutoff' and 'times'. If specified, removes all instances of 
+     'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
+     with airnow sites, setting 'group_var' to 'siteid' will remove all 
+     sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
+     only look at values at the beginning of each hour. Set 'times' to ''
+     if all times should be used.
```
rschwant, line 295:
```suggestion
     if all times should be used. This calculation occurs 
     over the entire analysis window and prior to calculating the regulatory metrics.
```

docs/appendix/yaml.rst
```diff
@@ -350,3 +357,13 @@ where domain_type is equal to domain_name.
...
+
+**data_proc:** This section stores all of the data processing information.
+
+   * **rem_obs_by_nan_pct:** Specify as dictionary with keys 'group_var', 
+     'pct_cutoff' and 'times'. If specified, removes all instances of 
+     'group_var' where there are > 'pct_cutoff' % NaN values. For example, 
+     with airnow sites, setting 'group_var' to 'siteid' will remove all 
+     sites with > pct_cutoff NaN values. Setting 'times' to 'hourly' will 
+     only look at values at the beginning of each hour. Set 'times' to ''
+     if all times should be used.
```
rschwant, line 369:
```suggestion
     if all times should be used. This calculation occurs 
     over the entire analysis window and prior to calculating the regulatory metrics.
```

rschwant (approved):
This looks great! If you agree, I just suggested a clarifying statement to the docs for how the calculation occurs. I tested it and it appears to work as I expect too.

Issue #146, opened 2022-12-01
Add filter option for AQS data specifically, but make it general


rschwant:
We need a general filter option. First we may want to use this for the AQS data to filter based on the measurement_scale to remove micro sites. But I'm sure there are other applications for this too, so we should make it general. This may be relevant for the aircraft evaluation too.

Maybe something like this in the yaml file:
- `filter_opt`: This could be the expr like '==' below
- `filter_dict`: This could be a dictionary where the keys are columns of the dataframe and the values are a list to accomodate multiple values being selected at once.

We probably just want to add into the code [a query option](https://pandas.pydata.org/docs/reference/api/pandas.DataFrame.query.html) like we have done with the domains. 

https://github.com/NOAA-CSL/MELODIES-MONET/blob/0d2ca12e5ec46f4f595ce3e18e1acd680a5534c0/melodies_monet/driver.py#L726-L728

This is just a start it probably would take a bit more time to think of a good format for this, but I don't think it would be too hard.

rschwant:
-Make sure keep the relevant site information from the observations.
-Filter on the number of observations or as a separate yaml structure.
-Two options: one prior to pairing and one during the analysis

mebruckner:
In order to do the filtering of Aeronet data I did for Brad Pierce's AGU22 poster, I add the following to my work flow between opening the observations and running the pairing.

```
# restriction of observations to analysis window specfied in yaml
an.obs['aeronet'].obj = an.obs['aeronet'].obj.sel(time=slice(an.start_time,an.end_time))

# removal of data for sites from the analysis scheme where there are fewer than x% of days reporting data
site_groups = an.obs['aeronet'].obj.groupby('siteid')#,'time'.dt.day)
for label,group in site_groups:
    #pull site, re-group to daily means
    tst = an.obs['aeronet'].obj['aod_550nm'][site_groups.groups[label]].groupby('time.dayofyear').mean().squeeze()
    if len(tst[~np.isnan(tst)]) < .5*len(tst): # filter to if <50% of days at site have data
        an.obs['aeronet'].obj['aod_550nm'][site_groups.groups[label]] = np.nan
        print('removed {}'.format(label))
        print('nobs:',len(tst[~np.isnan(tst)]))
```

rschwant:
Colin updated to add filtering options. So this is complete.

PR #156, opened 2023-01-10
Add filter option for observation data

colin-harkins:
Pull request to address issue #146.

This adds a yaml option to filter data more generally. The yaml option lives under the data_proc option in the plotting, statistics and obs sections of the yaml. 

An example of the formatting of this option is: `filter_dict: {'state_name':{'value':['CA','NY'],'oper':'isin'},'WS':{'value':1,'oper':'<'}}`
The above would filter data (airnow here) such that only observations in California or New york where windspeed is < 1 m/s remain.

When `filter_dict` is specified under `data_proc` in a plot group or in the statistics section, the filtering is performed right after the domain filtering. This is done with the pandas query function. 

To filter observations prior to pairing, a `data_proc` section with `filter_dict` specified can be added under the observation type. This will filter the observation data right after the mask and scale operations when reading in observations. This is done with the xarray Dataset.where function, because the observations are stored as an xarray Dataset and I could not figure out how to reproduce the same functionality with the xarray Dataset.query function.


zmoon (commented):
This is cool, but it seems like it would be simpler and more flexible to support passing in full query strings.

For example,
```
filter_dict: {'state_name':{'value':['CA','NY'],'oper':'isin'},'WS':{'value':1,'oper':'<'}}
```
could become
```yaml
filter: state_name in ['CA', 'NY'] and WS < 1
```
and this string passed to `.query()`.

colin-harkins:
I've updated this so that you can filter the data using a string as @zmoon suggested. This only applies to filtering in the plotting and statistics routines. 

The yaml option for this is `filter_string`. I've set it up such that if `filter_dict` is specified in the same `data_proc` section, `filter_string` will be ignored. Does that make most sense or would it be better to let the user specify both?

zmoon:
I think it is fine to expect one or the other, not both. And maybe raise error or warning if both are set.

colin-harkins:
I modified the behavior so that it just raises an error when both are set. I also changed the docs accordingly. I feel like that's probably safer than raising a warning that the user could potentially miss.

docs/appendix/yaml.rst
```diff
@@ -285,7 +300,25 @@ observation label is first and the model label is second
...
+     comparing to a list, "oper" can be set to "isin" or "isnotin" to filter 
+     by values in the list or not in the list, respectively. 
+     This cannot be specified if 'filter_string' is specified.
+     Example: {'state_name':{'oper':'isin','value':['CO']}, 
+     'WS':{'oper':'<','value':1}} 
+   * **filter_string:** This is a string used to filter the paired data sent 
+     to the plotting routine. The result is the same as using filter_dict.
+     This uses the pandas query method on the paired dataset.
+     This cannot be specified if 'filter_dict' is specified.
+     Example: state_name in ['CO'] and WS < 1
```
rschwant, line 321: Can we add something like the following?: "This option is only available for for surface and aircraft observations. For satellite observations, use the filter_dict option instead."

docs/appendix/yaml.rst
```diff
@@ -360,7 +393,25 @@ observation label is first and the model label is second
...
+     comparing to a list, "oper" can be set to "isin" or "isnotin" to filter 
+     by values in the list or not in the list, respectively. 
+     This cannot be specified if 'filter_string' is specified.
+     Example: {'state_name':{'oper':'isin','value':['CO']}, 
+     'WS':{'oper':'<','value':1}} 
+   * **filter_string:** This is a string used to filter the paired data sent 
+     to the statistics routine. The result is the same as using filter_dict.
+     This uses the pandas query method on the paired dataset.
+     This cannot be specified if 'filter_dict' is specified.
+     Example: state_name in ['CO'] and WS < 1
```
rschwant, line 414: Can we add something like the following?: "This option is only available for for surface and aircraft observations. For satellite observations, use the filter_dict option instead."

rschwant (commented):
Colin, this looks great! I tested it and it works well. I'm glad we added the filter_string option because you can do more complex filters that way like I wanted to test restricting WS < 1 m/s or WS is a NaN and with the filter_string option you can do this. I just suggested a couple small documentation updates and if you agree, I'll approve. Thanks!

colin-harkins:
@rschwant I've updated this with all of the suggestions you made.

rschwant (approved):
Great thanks, I approve, go ahead and merge it in.

PR #160, opened 2023-02-21
Add error when output_dir not specified

colin-harkins:
This pull request raises an exception when output_dir is not specified in the control .yaml file.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -514,8 +514,11 @@ class analysis:
         # set analysis time
         self.start_time = pd.Timestamp(self.control_dict['analysis']['start_time'])
         self.end_time = pd.Timestamp(self.control_dict['analysis']['end_time'])
-        self.output_dir = os.path.expandvars(
-            self.control_dict['analysis']['output_dir'])
+        if 'output_dir' in self.control_dict['analysis'].keys():
+            self.output_dir = os.path.expandvars(
+                    self.control_dict['analysis']['output_dir'])
+        else:
+            raise Exception('output_dir was not specified and is required. Please set output_dir in the control .yaml. ')
         if 'output_dir_save' in self.control_dict['analysis'].keys():
             self.output_dir_save = os.path.expandvars(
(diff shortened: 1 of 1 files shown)


melodies_monet/driver.py
```diff
@@ -514,8 +514,11 @@ def read_control(self, control=None):
         # set analysis time
         self.start_time = pd.Timestamp(self.control_dict['analysis']['start_time'])
         self.end_time = pd.Timestamp(self.control_dict['analysis']['end_time'])
-        self.output_dir = os.path.expandvars(
-            self.control_dict['analysis']['output_dir'])
+        if 'output_dir' in self.control_dict['analysis'].keys():
+            self.output_dir = os.path.expandvars(
+                    self.control_dict['analysis']['output_dir'])
+        else:
+            raise Exception('output_dir was not specified and is required. Please set output_dir in the control .yaml. ')
```
zmoon, line 521:
```suggestion
            raise Exception('output_dir was not specified and is required. Please set analysis.output_dir in the control file.')
```

zmoon:
Not sure what is causing the failure. The message says "NetCDF: Filter error: bad id or parameters or duplicate filter", but it was passing before and the versions of netcdf and xarray are the same.

colin-harkins:
It looks like it's failing with this message:

` FAILED melodies_monet/tests/test_get_data_cli.py::test_get_airnow_comp - subprocess.CalledProcessError: Command '['melodies-monet', 'get-airnow', '-s', '2019-09-01', '-e', '2019-09-02', '--dst', '/tmp/pytest-of-runner/pytest-0/test_get_airnow_comp0', '-o', 'x.nc', '--compress']' returned non-zero exit status 1.`

If I understand correctly, `test_get_airnow_comp` calls melodies_monet/_cli.py::get_airnow, which calls monetio's `airnow.add_data` function to download a small amount of airnow data from the web. I'm not sure if this is the case but if that airnow data hosted on the web has changed (in a way that would raise a netCDF error when saving to that format) or is unavailable for some reason, then that could be causing this to fail without any change on our end.

PR #162, opened 2023-02-23
Only set complevel for compressed float vars

zmoon:
We first noticed this in #160

In current netcdf-c (v4.9.1 when first noticed), it errors if you set compression for a variable-length dtype variable (such as string arrays with NumPy `object` dtype). With this PR, we only set compression for float vars.

See:
- https://github.com/Unidata/netcdf-c/pull/2231
- https://github.com/Unidata/netcdf4-python/issues/1205
- https://github.com/Unidata/netcdf-c/discussions/2554

---

* in `write_ncf`, only supply `complevel` for float vars
* add save/load to idealized example nb, which gets run in CI
* only supply `complevel` for float vars in `write_analysis_ncf`
* add some links/markup to the separate save/load example nbs


zmoon:
@colin-harkins we may have to do this in `write_analysis_ncf` as well. Maybe we can add a test.

zmoon:
@colin-harkins I think this is ready for you to take a look. I added some links and such to your example nbs.

PR #163, opened 2023-02-28
Add error when output_dir not specified (updated)

colin-harkins:
Opening a new pull request for pull #160 because I was having issues with getting it to recognize that the develop branch had been updated. This version incorporates the fixes from Zach that solve the tests failing.


zmoon (approved):
Looks fine to me!

PR #168, opened 2023-03-13
Add eval method for setting projection

colin-harkins:
Adds the ability to set projection in the control file with a string that will be evaluated using eval(). If the string begins with 'ccrs.', the eval method is triggered on the string.  For example `projection: ccrs.PlateCarre()` is the equivalent of m.proj = ccrs.PlateCarre().

### DIFF
--- a/docs/appendix/yaml.rst
+++ b/docs/appendix/yaml.rst
@@ -129,8 +129,32 @@ For example, ::
       PM25_TOT: 'PM2.5'
       O3: 'OZONE'
     
-**projection:** Not used currently in the code. This is under development and 
-likely to be moved to the "plots" section
+**projection:** In order to use the default projection for each model as defined 
+in the map_projection function in melodies_monet/plots/surfplots.py either remove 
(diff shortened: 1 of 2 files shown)


zmoon (approved):
Little dangerous to allow `eval` like that, even though checking that it starts with `ccrs.`. But I think it's fine given the way that MM is currently being used (you write your control file and run it yourself).

colin-harkins:
I updated the docs with a description based on what @rschwant wrote and the new functionality added here. 

@zmoon I agree and it crossed my mind as well but I came to the same conclusion about it.

rschwant:
@colin-harkins Just realized that your example on the docs is spelled wrong. 
https://melodies-monet.readthedocs.io/en/develop/appendix/yaml.html

Can you update:
For example, projection: ‘ccrs.PlateCarre()’.

to:
For example, projection: ‘ccrs.PlateCarree()’.

I tested that the latter one works. Thanks!

colin-harkins:
@rschwant good catch, I'll change it.

PR #172, opened 2023-03-29
Add ability to read aircraft obs from .csv file

colin-harkins:
This adds the ability to read aircraft obs from a .csv file. 

To accommodate different column names for the 'time' dimension and coordinates, I added the ability to set the column that should be used as time in the control file. This is set under the observation name as shown in the example below:

```
obs:
  sabre: # obs label
    filename: '/wrk/charkins/melodies_monet/aircraft/add_csv_read/SABRE-Merge_20230305_20230306_1719.csv'
    obs_type: aircraft
    time_var: TIME_START
    resample: '600S' 
```

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -120,6 +120,7 @@ class observation:
         self.data_proc = None
         self.variable_dict = None
         self.resample = None
+        self.time_var = None
 
     def __repr__(self):
         return (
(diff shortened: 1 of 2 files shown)


rschwant (approved):
This looks great, let's merge it in as an option. We can test it more before develop_aircraft goes to develop. Maggie can test it with SABRE and RAQMS output for example.

PR #176, opened 2023-04-10
Develop grid

dwfncar:
Part I of III for a possible approach to using the time chunking loop.  This part is for new code up to the 'pairing' stage, keeping datasets as xarray.


colin-harkins (commented):
My overarching comment is why does the `grid_data` option exist? I might have missed this being explained at some point.

For both opening obs and model data, this adds separate if statements for when the `grid_data` option is true vs when it is false. Based on my understanding, setting `grid_data: True` primarily allows the use of several things:
1. Reading files from `datadir` with filenames in the format of `filestr`, where placeholders such as YYYY and DD will be replaced with a date. This is done without any monetio readers.
2. The option to regrid the gridded data by setting `regrid: True`. 
3. Read hdf files for observations.

It seems like all of these things could be accomplished with slight modifications and additions to the respective functions for opening model and obs files (`melodies_monet.driver.model.open_model_files()` and `melodies_monet.driver.observation.open_obs()`) rather than using separate functions in `melodies_monet.util.read_grid_util`. Is there anything preventing that that I haven't noticed?

Otherwise I think the method of using the time chunking loop and regridding here looks fine, although I haven't been able to test this without data files.

melodies_monet/driver.py
```diff
@@ -394,6 +418,13 @@ def open_model_files(self, time_interval=None):
                 self.obj = mio.raqms.open_mfdataset(self.files,**self.mod_kwargs)
             else:
                 self.obj = mio.raqms.open_dataset(self.files,**self.mod_kwargs)
+        elif time_chunking_with_gridded_data:
+            date_str = time_interval[0].strftime('%Y-%m-%b-%d-%j')
+            print('model time chunk %s' % date_str)
+            model_datasets, filenames = read_grid_util.read_grid_models(
+                control_dict, date_str, model=self.label)
+            print(filenames)
+            self.obj = model_datasets[self.label]
```
colin-harkins, line 427: Wouldn't you want this if statement to occur prior to all of the other model options? I'm imagining if I were to try to use `time_chunking_with_gridded_data ` but `mod_type=='wrfchem'` or something else, the code would never make it to this `elif` statement.

If I'm understanding the purpose correctly, you would want to have this set up like follows (similar to in the `driver.observation.open_obs()` section):

````
if time_chunking_with_gridded_data:
    run some code...
else:
    if 'cmaq' in self.model.lower():
        run some code
    elif 'wrfchem' in self.model.lower():
        run some code
    ........

````

colin-harkins:
Besides potentially addressing my recent comment about the if statement for `time_chunking_with_gridded_data` in `driver.model.open_model_files()`, and addressing the failing tests, I think this looks good! I'll approve once I hear back about the if statement and once the tests pass.

colin-harkins (approved):

PR #178, opened 2023-05-19
Update resample_stratify function in tools.py to work with new stratify version

colin-harkins:
A new version of the stratify function was released to conda recently. This caused `melodies_monet.util.tools.resample_stratify()` to break due to the way it passes data to `stratify.interpolate()`. The error was 'DataArray' object has no attribute 'rechunk'. 

The new version of stratify assumes that the data passed is either as a `dask.array` or as a `numpy.ndarray`. Passing the data contained in the DataArray, rather than the DataArray itself fixes the issue.

Thanks to @quaz115 for bringing this error to my attention.

### DIFF
--- a/melodies_monet/util/tools.py
+++ b/melodies_monet/util/tools.py
@@ -276,7 +276,7 @@ def resample_stratify(da, levels, vertical, axis=1,interpolation='linear',extrap
     import stratify
     import xarray as xr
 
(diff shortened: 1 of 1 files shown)


quaz115:
The update works on my end, produces the .png plots and .csv for stats

rschwant:
Great, Quazi you can just review this code.

quaz115 (approved):
This updates works well, tested on my side.

PR #184, opened 2023-07-12
Fix CI

zmoon:
In the last few days started getting seg fault on some step of the tests.


zmoon:
Seems like it is the savefig test? Weird.

zmoon:
The linkcheck fail is some EPA websites, hopefully they will come back soon and can re-run.

zmoon:
Now it's only <https://www.epa.gov/aqs/> doing it. If not fixed by tomorrow afternoon I will merge anyway.

PR #201, opened 2023-10-23
Add pandas pin to install commands in docs

zmoon:
As discussed today, some MONETIO readers ([including](https://github.com/noaa-oar-arl/monetio/issues/111) AirNow and AERONET) and MM data processing routines[^1] don't work with pandas v2 yet.

[^1]: At least I thought I had a case of this at some point. Yes, just found one again, `df_mean=df.groupby(['siteid'],as_index=False).mean()` fails in spatial overlay plot.


rschwant (approved):
Looks great. Thanks for updating this so quickly.

PR #202, opened 2023-10-25
CLI: initial save/read paired dataset support for `melodies-monet run`

zmoon:
1. if `analysis.read` is present in the config, read saved paired datasets instead of pairing
2. if `analysis.save` is present in the config, save paired datasets

This is something I've been using for a while, and it's been working fine.

### DIFF
--- a/melodies_monet/_cli.py
+++ b/melodies_monet/_cli.py
@@ -146,7 +146,14 @@ def run(
         an.open_obs()
 
     with _timer("Pairing"):
-        an.pair_data()
+        if an.read is not None:
+            an.read_analysis()
+        else:
+            an.pair_data()
+
+    if an.save is not None:
+        with _timer("Saving paired datasets"):
+            an.save_analysis()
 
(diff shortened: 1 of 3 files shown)


melodies_monet/_cli.py
```diff
@@ -146,7 +146,14 @@ def run(
         an.open_obs()
 
     with _timer("Pairing"):
```
zmoon, line 148: Maybe should have separate message for each case here

zmoon: Oh but I guess the file names are already printed

zmoon:
linkcheck failed for <https://www3.epa.gov/ttn/naaqs/standards/ozone/data/20140131healthrea4a.pdf>, which is indeed currently unreachable

bbakernoaa:
@rschwant @colin-harkins can one of you take a look at this

colin-harkins (approved):
This looks good to me. Thanks for making a test for this.

PR #219, opened 2023-12-07
add multi-box plot user friendly version

btang1:
add multi-box plot user friendly version, already tested for 3 days without error in orion.

also update WRF-Chem example notebook with the new plots.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -903,7 +903,7 @@ class analysis:
             plot_type = grp_dict['type']
 
             #read-in special settings for multi-boxplot
-            if grp == 'plot_grp6':
+            if plot_type == 'multi_boxplot':
                 region_name = grp_dict['region_name'] #BEIMING
                 region_list = grp_dict['region_list'] #BEIMING
                 model_name_list = grp_dict['model_name_list']     
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -1198,9 +1198,6 @@ class analysis:
                                 #Clear info for next plot.
                                 del (comb_bx, label_bx, fig_dict, plot_dict, text_dict, obs_dict, obs_plot_dict)   
                         
-                        ####################################### 
-                        #This start BEIMING added multi-box-plot
-                        ########################################
                         elif plot_type.lower() == 'multi_boxplot':
                             if set_yaxis == True:
                                 if all(k in obs_plot_dict for k in ('vmin_plot', 'vmax_plot')):
@@ -1215,10 +1212,10 @@ class analysis:
                                 vmax = None
                             # First for p_index = 0 create the obs box plot data array.
                             if p_index == 0:
-                                comb_bx, label_bx,region_bx = splots.calculate_multi_boxplot(pairdf, pairdf_reg,region_name=region_name, column=obsvar, #BEIMING
+                                comb_bx, label_bx,region_bx = splots.calculate_multi_boxplot(pairdf, pairdf_reg,region_name=region_name, column=obsvar, 
                                                                              label=p.obs, plot_dict=obs_dict)
                             # Then add the models to this dataarray.
-                            comb_bx, label_bx,region_bx = splots.calculate_multi_boxplot(pairdf, pairdf_reg, region_name= region_name,column=modvar, label=p.model,  #BEIMING 
+                            comb_bx, label_bx,region_bx = splots.calculate_multi_boxplot(pairdf, pairdf_reg, region_name= region_name,column=modvar, label=p.model,   
                                                                          plot_dict=plot_dict, comb_bx=comb_bx,
                                                                          label_bx=label_bx)
 
@@ -1228,7 +1225,7 @@ class analysis:
                                 splots.make_multi_boxplot(
                                     comb_bx,
                                     label_bx,
-                                    region_bx,  #BEIMING 
+                                    region_bx,  
                                     region_list = region_list,
                                     model_name_list=model_name_list,
                                     ylabel=use_ylabel,
@@ -1243,10 +1240,7 @@ class analysis:
                                     debug=self.debug
                                 )
                                 #Clear info for next plot.
-                                del (comb_bx, label_bx,region_bx, fig_dict, plot_dict, text_dict, obs_dict, obs_plot_dict) #BEIMING
-                        ######################################
-                        #This end BEIMING added multi-box-plot
-                        ######################################
+                                del (comb_bx, label_bx,region_bx, fig_dict, plot_dict, text_dict, obs_dict, obs_plot_dict) 
 
                         elif plot_type.lower() == 'taylor':
                             if set_yaxis == True:
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -133,7 +133,7 @@ class observation:
             ")"
         )
 
-    def open_obs(self, time_interval=None):
+    def open_obs(self, time_interval=None, control_dict=None):
         """Open the observational data, store data in observation pair,
         and apply mask and scaling.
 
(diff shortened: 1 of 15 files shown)


melodies_monet/driver.py
```diff
@@ -902,6 +902,12 @@ def plotting(self):
             # get the plot type
             plot_type = grp_dict['type']
 
+            #read-in special settings for multi-boxplot
+            if grp == 'plot_grp6':
```
zmoon, line 906: This condition should use `plot_type` to check, since the plot group name can be anything.

melodies_monet/driver.py
```diff
@@ -902,6 +902,12 @@ def plotting(self):
...
+                region_name = grp_dict['region_name'] #BEIMING
+                region_list = grp_dict['region_list'] #BEIMING
```
zmoon, line 908: Maybe `domain_type`/`domain_name` and `domain_list` for more consistency with the other stuff?

melodies_monet/plots/scorecard.py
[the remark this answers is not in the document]
btang1: sure~

melodies_monet/driver.py
zmoon: For the user-facing part (e.g. the YAML), consistency is helpful.

btang1: the problem is yaml already have "domain_type" "domain_name",

so I added "region_name" and "region_list" and "model_name_list" to differentiate from previous.

so should we change to "domain_region_name" "domain_list" and "domain_model_name_list"?

dwfncar (approved):

melodies_monet/plots/surfplots.py
```diff
@@ -829,12 +830,168 @@ def calculate_boxplot(df, df_reg=None, column=None, label=None, plot_dict=None,
...
+
 def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, outname='plot',
                  domain_type=None, domain_name=None,
                  plot_dict=None, fig_dict=None,text_dict=None,debug=False):
+
+    """Creates box-plot. 
+
+    Parameters
+    ----------
+    comb_bx: dataframe
```
zmoon, line 899: 1. Needs to be a space on either side of the `:`
2. RHS should be the `type`, e.g. what you get if you used the `type()` function on the variable

```suggestion
    comb_bx : pandas.DataFrame
```

zmoon (changes requested):
@btang1

1. I see some docstring issues, I highlighted one. Not that important, but if you want to check how they render, you can [build the docs locally](https://melodies-monet.readthedocs.io/en/develop/develop/developers_guide.html#user-s-guide-development) and find [the page](https://melodies-monet.readthedocs.io/en/develop/api/melodies_monet.plots.surfplots.html) and see how it looks
2. Can you add multi-boxplot to the WRF-Chem--AirNow docs example as we discussed?
3. Can you go ahead and remove the BEIMING and such comments?
4. Can you update the "Description of All YAML Options" docs page with some info about this new plot type?

melodies_monet/plots/surfplots.py
```diff
@@ -932,6 +1123,10 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
     plt.tight_layout()
     savefig(outname + '.png', loc=4, logo_height=100)
 
+
+
+
```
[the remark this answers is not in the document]
btang1: Hey Zach, 

1) I can added updated yaml file and have a pull request to you for review.
2) I deleted all comments include 'BEIMING' and add some comments for plot inputs.

melodies_monet/plots/surfplots.py
btang1: thanks! I fixed that issue

melodies_monet/plots/surfplots.py
```diff
@@ -917,7 +1075,30 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
...
     sns.set_style("ticks")
-    sns.boxplot(ax=ax,x="variable", y="value",data=pd.melt(comb_bx), **boxplot_kwargs)
+    len_combx = len(comb_bx.columns)
+    if len_combx ==  3:
+        data_obs = comb_bx[comb_bx.columns[0]].to_frame().rename({comb_bx.columns[0]:'Value'},axis=1)
+        data_model1 = comb_bx[comb_bx.columns[1]].to_frame().rename({comb_bx.columns[1]:'Value'},axis=1)
+        data_model2 = comb_bx[comb_bx.columns[2]].to_frame().rename({comb_bx.columns[2]:'Value'},axis=1)
+        data_obs['model'] = model_name_list[0]
+        data_model1['model'] = model_name_list[1]
+        data_model2['model'] = model_name_list[2]
```
zmoon, line 1085: In theory there could be more than two models. It could be better to generalize these two `if` branches into one.

btang1:
Maybe late for 10 mins

Sent from my iPhone

On Dec 22, 2023, at 11:04 AM, Zachary Moon ***@***.***> wrote:

﻿

@zmoon commented on this pull request.

________________________________

In melodies_monet/plots/surfplots.py<https://github.com/NOAA-CSL/MELODIES-MONET/pull/219#discussion_r1435186924>:

> @@ -917,7 +1075,30 @@ def make_boxplot(comb_bx, label_bx, ylabel = None, vmin = None, vmax = None, out
                                'markersize': 20.0}}
     sns.set_style("whitegrid")
     sns.set_style("ticks")
-    sns.boxplot(ax=ax,x="variable", y="value",data=pd.melt(comb_bx), **boxplot_kwargs)
+    len_combx = len(comb_bx.columns)
+    if len_combx ==  3:
+        data_obs = comb_bx[comb_bx.columns[0]].to_frame().rename({comb_bx.columns[0]:'Value'},axis=1)
+        data_model1 = comb_bx[comb_bx.columns[1]].to_frame().rename({comb_bx.columns[1]:'Value'},axis=1)
+        data_model2 = comb_bx[comb_bx.columns[2]].to_frame().rename({comb_bx.columns[2]:'Value'},axis=1)
+        data_obs['model'] = model_name_list[0]
+        data_model1['model'] = model_name_list[1]
+        data_model2['model'] = model_name_list[2]


In theory there could be more than two models. It could be better to generalize these two if branches into one.

melodies_monet/plots/surfplots.py
zmoon: now is loop

zmoon:
@rschwant can you give a quick check to the YAML documentation updates here?

rschwant (commented):
This is great to have this feature added!

Some thoughts:

1) Do you actually need a different submodule for these: calculate_boxplot and calculate_multi_boxplot? The plotting routine is also quite similar, but I agree with how you have it to make it separate routines to avoid confusion. For the calculations, it just looks like you add the region name for each iteration, but this could also easily be done by adding an option in calculate_boxplot that says multi = False by default and then you set multi to True when using it for multi-boxplot calculations and then you add (region_bx = pd.DataFrame() and region_bx['set_regions']=df[region_name[0]] ) when multi = True? Currently, you only add region_bx['set_regions']=df[region_name[0]] this line in the calculations for non regulatory calculations, why? I do not see why it needs to be restricted in this way? And having it applied for both reg and non-reg variables makes sense. If this plot only works for non-regulatory calculations, we should describe this on ReadTheDocs.

2) For the yaml updates here:
https://melodies-monet.readthedocs.io/en/latest/appendix/yaml.html
Under the first category called General Rules, I have a list of plots for how the sampling window is applied. Can you add a line for "multi-boxplot" here and describe how the sampling window works, likely the same as box-plot?

3) Can you update the code in the ReadTheDocs to include this as an option under Supported Plots including an example figure. This can be on any dataset you don't have to choose the standard one.
https://melodies-monet.readthedocs.io/en/latest/background/supported_plots.html

4) In the calculate_multi_boxplot comments, you refer still to box-plot several times. If you keep this submodule, can you update this to read multi-boxplot? And likewise for make_multi_boxplot, you refer to box-plot and calculate_boxplot several times, can you update these to multi-boxplot and calculate_multi_boxplot (if you keep this). 

Thanks for adding this feature and let me know if you have questions!

rschwant:
Actually for number 1, I just realized that you still have different outputs, so you have to keep these separate even though they are largely similar. So actually, then I just have the regulatory option question for number 1.

btang1:
Added regulatory feature for all-region-multi-boxplot. and tested for 15 days for ozone and pm2.5.
result can be accessed in slide: https://docs.google.com/presentation/d/1AyaUUOqrDEfnyO5sPaiQq-oiaieYPuaP/edit?usp=sharing&ouid=117233739375959098251&rtpof=true&sd=true

melodies_monet/plots/surfplots.py
```diff
@@ -880,6 +881,13 @@ def calculate_multi_boxplot(df, df_reg=None, region_name= None,column=None, labe
     plot_kwargs['label'] = label
     if df_reg is not None:
         comb_bx[label] = df_reg[column+'_reg']
+        siteid_bx['siteid']=df_reg['siteid']
+        region_list = []
+        for i in range(len(siteid_bx['siteid'])):
+            siteid_here = siteid_bx['siteid'][i]
+            region_here = df[df.siteid == siteid_here][region_name[0]].to_string().replace(' ',',').split(',')[-1]
+            region_list.append(region_here)
+        region_bx = pd.DataFrame(region_list,columns=['set_regions'])
```
rschwant, line 890: @zmoon This looks broadly good. Zach, there is probably an easier way to do this with either merge or join in pandas. The metadata is only in df and not in df_reg so we just need to add it into df_reg here by siteid. Can you advise Beiming on how to do this?

btang1: @zmoon shall we merge this PR?

btang1:
Hi @zmoon @rschwant, I have updated regulatory with pandas.merge and tested for 15 days. now it is more concise compared to previous one. I will update a slide for detail result. please review and if no problem. let's merge it. Thanks!

btang1:
@zmoon @rschwant link to the slide: https://docs.google.com/presentation/d/1bXBJy82KdvDELJroXbw62ZIFLAaM3Azh/edit?usp=share_link&ouid=117233739375959098251&rtpof=true&sd=true

rschwant (approved):
Thanks for making the code cleaner. I approve! Later on we could generalize this to include more than just the EPA_Region, but this could be done in another step.

btang1:
> Thanks for making the code cleaner. I approve! Later on we could generalize this to include more than just the EPA_Region, but this could be done in another step.

Thanks @rschwant now shall we merge this PR request?

rschwant:
Looks like Zach has requested a change, but likely you addressed it?, so as long as @zmoon you approve we can merge this in.

melodies_monet/plots/surfplots.py
zmoon: I believe the 169facb change captured what we were looking for.

rschwant: Great, I agree.

melodies_monet/plots/surfplots.py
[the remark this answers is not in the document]
zmoon: This comment was in regards to the added blank lines. Two blank lines between function definitions is enough.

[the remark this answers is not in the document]
btang1: I think I correct this. no space in new version.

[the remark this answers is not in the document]
zmoon: Thanks

zmoon (approved):

PR #246, opened 2024-02-12
Add function to loop over pairings, saving out netcdf files for each

colin-harkins:
This adds a function to loop over pairings, saving a paired netcdf file for each set. 

This is currently intended for aircraft but I wrote it in a way that it should work for other types of pairings as well. The best way I could think of to do this was to make the user always provide a mapping between the obs and corresponding model files. This avoids having to do any complicated logic of finding the right model files for each obs file, which might need to be written specific to each model.

Currently the mapping between obs and model for each pairing is provided as a dict to the function but I'm going to expand this so that it can be provided as a .yaml file. There is also another function to read a list of obs files and then give back the time bounds of each, so that it is easier for the user to assemble matching model files.

This is an example of the mapping with some files I had on hand: 

```
file_pairs = {'0722':{'model':{'wrfchem_v4.2':'/wrk/users/charkins/melodies-monet_data/wrfchem/run_CONUS_fv19_BEIS_1.0xISO_RACM_v4.2.2_racm_berk_vcp_noI_phot/0722/*'},
                      'obs':{'firexaq':'/wrk/d2/rschwantes/obs/firex-aq/R1/10s_merge/firexaq-mrg10-dc8_merge_20190722_R1.ict'}},
            '0905':{'model':{'wrfchem_v4.2':'/wrk/users/charkins/melodies-monet_data/wrfchem/run_CONUS_fv19_BEIS_1.0xISO_RACM_v4.2.2_racm_berk_vcp_noI_phot_soa/0905/*'},
                    'obs':{'firexaq':'/wrk/d2/rschwantes/obs/firex-aq/R1/10s_merge/firexaq-mrg10-dc8_merge_20190905_R1.ict'}}
```

Putting this as a draft PR right now because I'm still cleaning this up and I'm very much open to comments on whether this addresses all of the functionality necessary for aircraft time looping or if this should be done another way.


colin-harkins:
Closed this to first merge develop into develop_aircraft. After that is done I'll open a new pull request for this.

PR #248, opened 2024-02-13
Merge develop branch into develop_aircraft and fix several merge conflicts

colin-harkins:
This pull request merges everything up to present from the develop branch into develop_aircraft. There were also a couple merge conflicts that needed to be resolved included here, as well as some small changes to a few if statements that were originally specific to `obs_type == 'pt_sfc'` in develop but actually are currently meant to function for everything except satellite obs.


colin-harkins:
@quaz115 if you're able to just confirm this functions as expected for you with the aircraft and surface obs then I'll proceed.

quaz115:
> @quaz115 if you're able to just confirm this functions as expected for you with the aircraft and surface obs then I'll proceed.

@colin-harkins _It generates all aircraft plots in develop_aircraft branch as expected_ (**See plots on Meiyu at:** `/wrk/qrasool/output_testaircraft_PR189_uptoviolin_multiplemodels`), however would like to point the few warnings i get :
These warnings don't effect the plots being generated though.

1) for ```
#This just pairs the data
an.pair_data():
```
I see this warning:
```

/home/qrasool/MM_developaircraftPR248/MELODIES-MONET/melodies_monet/util/tools.py:313: FutureWarning: A value is trying to be set on a copy of a DataFrame or Series through chained assignment using an inplace method.
The behavior will change in pandas 3.0. This inplace method will never work because the intermediate object on which we are setting values always behaves as a copy.

For example, when doing 'df[col].method(value, inplace=True)', try using 'df.method({col: value}, inplace=True)' or df[col] = df[col].method(value) instead, to perform the operation inplace on the original object.


  df_model.pressure_model.fillna(df_model.pressure_obs,inplace=True)
```

2)  For `an.plotting()` while generating plots it gives this warning (repeated for each plot being generated):
```
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
Warning: set rem_obs_nan = True for regulatory metrics
```

3) also boxplot related seaborn FutureWarning:

```
/home/qrasool/MM_developaircraftPR248/MELODIES-MONET/melodies_monet/plots/surfplots.py:978: FutureWarning: 

Passing `palette` without assigning `hue` is deprecated and will be removed in v0.14.0. Assign the `x` variable to `hue` and set `legend=False` for the same effect.

  sns.boxplot(ax=ax,x="variable", y="value",data=pd.melt(comb_bx), **boxplot_kwargs)
```

colin-harkins:
@quaz115 Thanks for pointing out these warnings. I'll fix both these FutureWarnings. I need to update the other packages in my environment to have them come up but that won't take long. I'm not sure about the cause of the `set rem_obs_nan = True` one but let me look into it and see. I think it's coming from the code in Develop. Let me see if it's a warning that should be coming up every time or if it's a problem in the logic causing it to come up.

PR #249, opened 2024-02-13
CLI: initial AQS support

zmoon:
Some things in monetio are not working currently, like network selection and other aspects involving the separate metadata. But getting hourly and daily data seem to be working ok here.


melodies_monet/_cli.py
```diff
@@ -914,6 +915,221 @@ def get_utc_offset(*, lat, lon):
...
+@app.command()
+def get_aqs(
+    start_date: str = typer.Option(..., "-s", "--start-date", help=f"Start date. {_DATE_FMT_NOTE}"),
+    end_date: str = typer.Option(..., "-e", "--end-date", help=f"End date. {_DATE_FMT_NOTE} {_DATE_END_NOTE}"),
+    daily: bool = typer.Option(False, help=(
+            "Whether to retrieve the daily averaged data product. "
+            "By default, the hourly data is fetched."
+        )
+    ),
+    param: List[str] = typer.Option(["SPEC", "PM2.5", "PM10", "VOC"], "-p", "--params", help=(
```
zmoon, line 927: @rschwant @bbakernoaa what do you think defaults should be?

zmoon: @rschwant suggested default could be O3, PM2.5, PM10

rschwant: Yes, I agree. I'm testing this out right now too.

rschwant:
@zmoon, how would you use this code to download the daily data? I tried this: 
melodies-monet get-aqs -s 2021-01-01 -e 2021-01-03 --daily but I get this error: 
Fetching data with monetio failed in 146 seconds
Error message: 'time'

zmoon:
@rschwant I'll look into that. Did you try without --daily?

rschwant:
Yes, without daily works, but I want to try to pull the daily averaged data from IMPROVE.

zmoon:
Ok, I can do a monetio hotfix for it.

melodies_monet/_cli.py
zmoon: Updated in e54849f , with some additional help text.

zmoon:
CI failure is related to the NCDC website the ISH files are on failing.

zmoon:
@rschwant daily option should be working with this branch of monetio: https://github.com/noaa-oar-arl/monetio/pull/163 I'll also comment here when it has been released via conda-forge (hopefully soon).

rschwant:
The daily option works with your updates now. Thanks for updating. Would it be possible to extend the default meta data to include some additional options for both daily and non-daily include the following: method_code, method_name, networks, monitor_type, measurement_scale, measurement_scale_definition, land_use, & epa_region. People may want to filter and group on these options in the future.

zmoon:
@rschwant method code and name are available without the AQS metadata loading, but not currently usable, since the long-to-wide format transformation doesn't (currently) create separate columns for them for each parameter, so currently I am skipping them:

https://github.com/NOAA-CSL/MELODIES-MONET/blob/7b9406b8c082e52cf548de521156122a12a2660c/melodies_monet/_cli.py#L1058

The others [will come](https://github.com/zmoon/monetio/blob/4ef1028c179d333650387225178c9b9e54d9313b/monetio/obs/epa_util.py#L524) from the AQS metadata files, once that is fixed.

So, in the future, after more monetio work, we should be able to get all of those that you mentioned. But we might as well get this PR in first, since it at least gets the data.

melodies_monet/_cli.py
```diff
@@ -914,6 +915,250 @@ def get_utc_offset(*, lat, lon):
...
+@app.command()
+def get_aqs(
+    start_date: str = typer.Option(..., "-s", "--start-date", help=f"Start date. {_DATE_FMT_NOTE}"),
+    end_date: str = typer.Option(..., "-e", "--end-date", help=f"End date. {_DATE_FMT_NOTE} {_DATE_END_NOTE}"),
+    daily: bool = typer.Option(False, help=(
+            "Whether to retrieve the daily averaged data product. "
+            "By default, the hourly data is fetched."
+        )
+    ),
+    param: List[str] = typer.Option(["O3", "PM2.5", "PM10"], "-p", "--params", help=(
```
zmoon, line 927:
```suggestion
    param: List[str] = typer.Option(["O3", "PM2.5", "PM10"], "-p", "--param", help=(
```

rschwant:
Sounds we can add this later then. Let me know when this is ready to go in then and I can approve.

melodies_monet/_cli.py
```diff
@@ -914,6 +915,250 @@ def get_utc_offset(*, lat, lon):
...
+    ),
+    num_workers: int = typer.Option(1, "-n", "--num-workers", help="Number of download workers."),
+    verbose: bool = typer.Option(False),
+    debug: bool = typer.Option(
+        False, "--debug/", help="Print more messages (including full tracebacks)."
+    ),
+):
+    """Download EPA AQS data using monetio and reformat for MM usage.
+
+    These are archived data, stored in per-year per-parameter files, described at
```
zmoon, line 960:
```suggestion
    These are archived data, stored in per-year per-parameter-group files, described at
```

zmoon:
@rschwant final look?

rschwant (approved):
Looks great, I approve!

PR #250, opened 2024-02-15
Add tool to loop over sets of pairings, saving out a netcdf for each

colin-harkins:
This adds a function to loop over pairings, saving a paired netcdf file for each set.

This is currently intended for aircraft but I wrote it in a way that it should work for other types of pairings as well. The best way I could think of to do this was to make the user always provide a mapping between the obs and corresponding model files. This avoids having to do any complicated logic of finding the right model files for each obs file, which might need to be written specific to each model.

Currently the mapping between obs and model files for each pairing is provided as either a dict or a separate yaml file.

There is also another function to read a list of obs files and then give back the time bounds of each, so that it is easier for the user to assemble matching model files.


colin-harkins:
One thing that I wanted to get feedback on is the handling of variable names with some special characters from the obs dataset. For example, in the firexaq R1 data, there are variable names that contain `/` and `&`, neither of which can be used in a netcdf variable name. Currently I just drop any variables with those characters when saving but that isn't preferable. I could change it to error out when this happens or give a warning when certain variables have been dropped. I think that the user could provide a yaml that renames specific obs variables so that this doesn't happen once they've been warned about it.

rschwant:
@colin-harkins, is it possible to add an attribute to netcdf like how the units are added named something like long_name. Then in all cases just repeat the variable name in the long_name field. And for any characters that are not allowed in netcdf variable names, can you just drop those characters in the variable name, but they would still be listed in the long_name for reference later on. I've seen this done in the ATom netcdf files and this should work well for our purposes and we can explain this in our documentation. I feel like other items are going to fail if we keep characters like this in the variable names too like titles of plots, so best just to drop characters like this now. What do you think?

colin-harkins:
@rschwant I think that is a great solution. I'll implement that today and update the PR.

quaz115:
@colin-harkins This PR works well for develop_aircraft plots.
Just wanted to ask that did you remove this yaml file intentionally in this PR: ` loop_pairing(control=control_fn,file_pairs_yaml='**aircraft_looping_file_pairs.yaml**') `
Since, this is in the examples/jupyter_notebooks/Monet-example_aircraft_pairing_loop.ipynb but doesn't work as the yaml file isn't present.  (if this feature doesn;t work anymore you can remove it in the notebook)
Beside the above clarification , i am going ahead to approve the PR as i can see the pairing netcdf files for different files created with variable names that contain / and & ,such as: Cl_ug/m3_DIBB' being present.

rschwant (commented):
Looks great, thanks Colin for working on this!

Just a couple small questions and updates:

1) Can you provide an example of the format of this file? aircraft_looping_file_pairs.yaml I see this in the comments in the code which is great, but an example would be useful too.

2) When you provide a control_fn to do this looping? Are all the fields specified in this example (examples/yaml/control_aircraft_looping.yaml) used? For example, the analysis time is the following:
start_time: '2019-09-05-12:00:00' #UTC
end_time: '2019-09-06-00:00:00' #UTC
Does this need to be the full range of the data in the obs and model? Or is this not used? I wonder if just like how you overwrite the model and obs files, you have to add the analysis window here too?

3) The jupyter notebook is great for this, but if we are running for the entire campaign and especially if we resample at 1 minute intervals, we will need to submit this as a job to hera. Could you include an example python script for this too? You can add it into this folder: https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/submit_jobs/. This would just be a .py file for the python commands in your jupyter notebook file. And I would just default to using a yaml file to input the file pairs here. This would use the same .sh script that is already there to call the python file, so you don’t need to provide another example for submitting the python script itself. 

4) Can you provide an example, of how then you would read in and combine these netcdf files to process the data together. No need to show this for all the plot types. Maybe you can just show this for the timeseries aircraft plot because this is a good test that all is working. Most of the netcdf data will be the same, but the analysis window may be different, so I'm trying to figure out what analysis window would be used when pulling these paired datasets together?

colin-harkins:
I made changes to address the comments from @rschwant  and @quaz115.

1) aircraft_looping_file_pairs.yaml is now included in a subfolder for supplementary yaml files

2) When providing a control file for the looping, many of the fields I previously had in the file are not necessary. I've commented those out in the updated version. For start_time and end_time, the fields are not necessary for the looping and are only necessary for plotting and one of the satellite pairing options. To reduce confusion, I made start_time and end_time optional in the control yaml file. Let me know what you think about this, but I don't think it will cause issues to make them optional.

3) I provided a .py file in the submit_jobs folder that I've tested on hera with this code

4) I provided a jupyter notebook and control yaml that demonstrate reading in the sets of paired observations. It works pretty much the same as the read analysis examples, but uses the netcdf files generated by the aircraft pairing loop. The only issue here is reading in model and obs data. Currently there is no functionality for reading in multiple icartt files, so the control file I made just uses a single day. This doesn't really affect the example and I can work on adding the functionality for reading multiple icartt files in the same control file in the future or I can try to add that to this PR as well. The jupyter notebook provided also makes a timeseries plot as a basic demonstration that this is working. 

I still need to come up with a way to handle some of the plotting that requires `analysis.models` and `analysis.obs` directly rather than just `analysis.paired`, because reading in an entire campaigns worth of model data when plotting is just not really feasible in some cases. I think that is something that can be addressed in the future once develop_aircraft is merged into develop.

If this looks good to you, I'll merge it in and then start a PR for develop_aircraft into the develop branch.

rschwant (approved):
This looks great thanks Colin! I think adding the analysis time as optional is okay here as it is then clearer when this is used and that it is needed for the plotting, but not the pairing step. We can work on getting the plots that use the obs or model output figured out at another step, so I will make a note that this looping only works for plots that use the paired class.

colin-harkins:
@rschwant Thanks for looking at this. I'll merge and make a pull request into develop today.

PR #253, opened 2024-03-18
Merge develop_aircraft into develop

colin-harkins:
This pull request merges the develop_aircraft branch into the develop branch. Nothing here has changed from #246.


colin-harkins:
Looks like I need to fix something to get the tests to pass. I'll update today.

colin-harkins:
I couldn't quite figure out how to get the docstring for the loop_pairing function to work when providing examples of the inputs, so i just gave locations where examples can be found for now. Unless you have any recommendations right now, I can try to figure out a way to do this in the future once I can get more familiar with the syntax of sphinx docstrings.

rschwant:
@colin-harkins updating to showing an example in jupyter notebook works for the loop_pairing function. Is this ready then? I will look through it early this week if so, and @quaz115 if you can double check it all looks good too that would be great. We need 2 reviewers to go into develop.

quaz115:
> I couldn't quite figure out how to get the docstring for the loop_pairing function to work when providing examples of the inputs, so i just gave locations where examples can be found for now. Unless you have any recommendations right now, I can try to figure out a way to do this in the future once I can get more familiar with the syntax of sphinx docstrings.

@colin-harkins @rschwant  Regarding the docstring i thinks the line at which sphinx-build error happens is causing the issue:  i.e.    `An example can be found in examples/jupyter_notebooks/Monet-example_aircraft_pairing_loop.` 

You can try with closing the file path with backticks like this ? 

`An example can be found in `examples/jupyter_notebooks/Monet-example_aircraft_pairing_loop.ipynb``

**Hoping the above fix might be sufficient, if not also look at the following suggestion (since Sphinx-build failed at line right before where file paths have asterisks as well):**

Sphinx seems to cause issue with CI/CD tests while reading file paths specially when asterisks (*) in the file paths or in text. Check for other instances as well where there is need to enclose the file paths in double backticks (which is used for inline code in reStructuredText i.e. docstring). 

For example: you can also try changing current format :

`file_pairs_yaml = 'path/to/some/file*.yaml'`

**TO**

Either: 
file_pairs_yaml = 'path/to/some/file\*.yaml'  # If you are using asterisks literally. (but that is not the case as they are used as wildcards for reading multiple files)

**OR** a better alternative would be to simply highlight the text without any misunderstanding by Sphinx:

Example:
`file_pairs_yaml = path/to/some/file*.yaml`  # Using backticks to enclose paths or commands.

rschwant:
@colin-harkins Would it be better to merge your commits from colin-harkins:develop_aircraft to NOAA-CSL:develop_aircraft and then do a merge from NOAA-CSL:develop_aircraft to NOAA-CSL:develop, or does this not matter? I want to close out the develop_aircraft branch after this merge.

colin-harkins:
@rschwant I don't think it really matters since we want to close out the develop_aircraft branch after this. That's what I was thinking when submitting this pull request anyway.

rschwant (approved):
Sounds good, we can keep it coming from your fork then, since I double checked your fork is updated with the latest develop_aircraft branch and we are deleting this branch after the merge anyway. Everything looks good to me. Quazi will be the second approver and he may not get to this until next Monday.

colin-harkins:
@quaz115 I fixed the issue with the docstring by using a literal block and then also enclosed the file paths in double backticks and everything seems to work. Thanks for the help with that.

quaz115:
> @quaz115 I fixed the issue with the docstring by using a literal block and then also enclosed the file paths in double backticks and everything seems to work. Thanks for the help with that.

@colin-harkins Awesome! Glad it worked. I will try to review the PR by today or tomorrow and then use this to submit a new PR with the remaining aircraft vs model plots (curtain and spatial)

quaz115 (approved):
Reviewed both the aircraft plotting and netcdf pairing updates and tested as well, they work properly and have the most recent code changes.

PR #255, opened 2024-03-27
Add variable summing and PM unit conversions for firexaq dataset

colin-harkins:
This pull request adds the functionality for summing up variables into a new variable for models and obs. An example of how this is used is that model PM variables may need to be summed when comparing to observational variables for PM. 

Additionally, this pull requests adds functions for some of the unit conversions between std and amb needed to compare with variables in firexaq.

This is developed based on the code attached below.

[analysis_LA_both_aerosols_vert_mix_v422_phot.txt](https://github.com/NOAA-CSL/MELODIES-MONET/files/14781987/analysis_LA_both_aerosols_vert_mix_v422_phot.txt)

### DIFF
--- a/examples/yaml/control_wrfchem_aircraft_sumpmspecies.yaml
+++ b/examples/yaml/control_wrfchem_aircraft_sumpmspecies.yaml
@@ -1,0 +1,273 @@
+# General Description:  
+# Any key that is specific for a plot type will begin with ts for timeseries, ty for taylor
+# Opt: Specifying the variable or variable group is optional
+# For now all plots except time series average over the analysis window. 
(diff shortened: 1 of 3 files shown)


rschwant:
@colin-harkins 
This looks great Colin! I just have a couple very small changes.

I realized that the units in your example for the BC are in ng instead of ug. So can you add the following change in the example yaml file: 

'BC_mass_90_550_nm_SCHWARZ':
        unit_scale: **1000.**  #Opt Scaling factor 
        unit_scale_method: '**/**' #Opt Multiply = '*' , Add = '+', subtract = '-', divide = '/'

And then I also realized that the example WRF-Chem file you and Quazi are using is an intermediate WRF-Chem simulation of the RACM2B-VCP mechanism that has very low OA and maybe some other problems too. Best to use a more standard file, so can you update to use these files instead (I believe these are the vertical data for the surface data that is now on the ftp site for the auto examples in MELODIES MONET): /wrk/d2/rschwantes/wrf/firex_mech/wrfchem_example/racm_esrl/*

Then the organic aerosol much more closely aligns with what my previous script got, which is a great QC of all of this. I tested these updates and they should work well, but let me know if you run into problems.

Everything else looks great!

colin-harkins:
@rschwant I made the changes you requested, if you're able to approve this I'll merge it in.

rschwant (approved):
Looks great, feel free to merge in when you have the chance. Thanks!

PR #257, opened 2024-05-01
Update data CLI doc

zmoon:
Missed adding AQS to the list in #249 

Also updated the supported-datasets page a bit


rschwant (approved):
Looks good, thanks for updating!

PR #260, opened 2024-06-23
subset_MODIS_l2

dwfncar:
Added subset_MODIS_l2 to util/time_interval_subset.py.


dwfncar:
Currently working through a bug in the grid_utils.py module, other workflow choices at the driver/script level.
Will reopen PR once these are resolved.

PR #268, opened 2024-08-20
fix bug that removed aircraftplots import line

rschwant:
This line got accidentally removed during one of the satellite evaluation merges, so let's add it back in so that the aircraft plots are properly imported.


zmoon (approved):
Looks good to me

PR #270, opened 2024-09-11
CLI: `get-openaq`

zmoon:
Note that this uses https://github.com/noaa-oar-arl/monetio/pull/141 (merged into monetio develop 2024-Sep-19)

* [x] add a test or two
* [x] add reference-grade / low-cost flags


zmoon:
@jordanschnell , example one day command for you to try:

```
melodies-monet get-openaq -s 2024-09-10 -e '2024-09-10 23:00' -p o3 -p pm25 -p pm10 -p co -p no2 --verbose --debug -n 4
```

If you don't already have one, [register for OpenAQ API key](https://docs.openaq.org/docs/getting-started) and then set the environment variable `OPENAQ_API_KEY` and the tool will use it.

zmoon:
@jordanschnell The reader prints some debug logging messages related to the API calls that should shed some light into how much it is timing out (you can see the individual web API calls and when they get rate limited). I've set up an env var you can use to enable this: `MM_LOGGING_LEVEL` (use `debug` or `info`).

zmoon:
I added individual flags for the two sensor types (reference-grade and low-cost). But this could also be done like the params are, with a type flag.

jordanschnell (commented):
All is working well - approved.

melodies_monet/_cli.py
```diff
@@ -1160,6 +1165,246 @@ def get_aqs(
...
+            "Output file name (or full/relative path). "
+            "By default the name is generated like 'OpenAQ_<start-date>_<end-date>.nc'."
+        )
+    ),
+    dst: Path = typer.Option(".", "-d", "--dst", help=(
+            "Destination directory (to control output location "
+            "if using default output file name)."
+        )
+    ),
+    param: List[str] = typer.Option(["o3", "pm25", "pm10"], "-p", "--params", help=(
```
zmoon, line 1183: I think makes more sense singular, unlike AQS and its parameter groups.
```suggestion
    param: List[str] = typer.Option(["o3", "pm25", "pm10"], "-p", "--param", help=(
```

melodies_monet/_cli.py
```diff
@@ -1160,6 +1165,246 @@ def get_aqs(
...
+        typer.secho(f"Error: method {method!r} not recognized", fg=ERROR_COLOR)
+        raise typer.Exit(2)
+
+    start_date = pd.Timestamp(start_date)
+    end_date = pd.Timestamp(end_date)
+
+    if method in {"openaq-fetches"}:
+        dates = pd.date_range(start_date, end_date, freq="D")
+    elif method in {"api-v2"}:
+        dates = pd.date_range(start_date, end_date, freq="H")
```
zmoon, line 1233: The `'H'` is deprecated
```suggestion
        dates = pd.date_range(start_date, end_date, freq="h")
```

PR #273, opened 2024-09-20
add camx as model

blychs:
I just realized that I forgot to do this.
This just adds the CAMx reader as a possible model. It has been in MONETIO for a while now, and I have been using it in my fork, but (stupidly) never committed it to the main repo.
Cheers
Pablo


blychs:
I am not sure about who I should at. @zmoon ? @rschwant ?

PR #274, opened 2024-09-20
Add camx as possible model

blychs:
I just realized that I forgot to do this.
This just adds the CAMx reader as a possible model. It has been in MONETIO for a while now, and I have been using it in my fork, but (stupidly) never committed it to the main repo.
Cheers
Pablo

PD: I don't know who I should @. @rschwant ? @zmoon ?


blychs:
I am sorry for all the noise, I updated things into the wrong branch. If I did it correctly, there should only be 3 lines of changes (in the driver) to add CAMx

PR #276, opened 2024-09-23
Swath obs gridding and MODIS L2 example

dwfncar:
MODIS L2 swath data example L2 observations and model gridded to an intermediate observation grid.


rrbuchholz (approved):
Hi David,
It looks good to me. The only concerns could be in the driver, but those additions are protected regrid if statement, only used in this workflow, so it should be fine.
Just need to add the comments that are missing and it is good to merge.
Cheers,
Rebecca

Issue #278, opened 2024-09-24
Region support only works with obs Dataframes


blychs:
Dear all,
Currently, the region support that works with AirNow does not work for everything else the way it should.
The code is reading regions from the information already provided by the obs DataFrame (the driver does a query. However, with EPA and Giorgi regions there is no reason for that: we have boxes defined in MONET, that MONETIO can import.
The sat data with the current setup just wouldn't work for any domain_type different from 'all' (I don't know about aircraft).
If @rschwant and @rrbuchholz agree, I suggest changing it (for now, I could stick to the satdata for that) so that it can subset the domain, at least in the case of EPA and Giorgi regions. Adding arbitrary square boxes should be close to no extra effort.
We can mention this in the tutorial as WIP.
Let me know your thoughts
Pablo

rschwant:
The domains being based on a query is still useful though. Because the surface observations normally have meta data to separate out the domains, so I do not want to change this feature. As you can easily plot based on state, city, etc. this way. The boxes in MONET and MONETIO I doubt are as accurate as they would be based on lat lon boxes that may not perfectly represent the domain. But adding the capability to use domains for data that does not provide meta data for this would be useful. But we need to do it in a way that does not break the query feature for surface observations. Can we add a yaml true/false statement for this? Something like domain_query = True (default and domains are based on the meta data in the observation file) and if you set this to False, then it uses the boxes defined in MONET and MONETIO to do this query. What do you think about this?

blychs:
That is a good solution. I am happy to move forward with that, or at least adding an extra check for the satdata.

rschwant:
Great thanks!

blychs:
Closing this as it has been solved in #280

blychs:
Closing this as it has been solved in #280

Issue #272, opened 2024-09-20
Add (optional) arbitrary region support with regionmask


blychs:
Dear all,
If we allow people, as an optional dependency, to include the package "regionmask", we could easily allow for using arbitrary shapefiles, counties, etc.
https://regionmask.readthedocs.io/en/stable/
If you are interested, I should be able to add support for it.
Cheers
Pablo

rschwant:
Yes, sorry this got lost in my emails. Please proceed forward with this. 

We do have filtering options already, so just make sure to consider these filtering options while you are coding this in.
https://melodies-monet.readthedocs.io/en/stable/appendix/yaml.html (search for filter_dict) which can be applied in the observations or in the data_proc section for plotting and statistical calculations.

blychs:
Will be fixed by #326 when merged.

rschwant:
This is done as #326 is merged.

PR #280, opened 2024-09-25
Add domain_query option

blychs:
Solves Issue #278 .
As suggested by @rschwant , adds optional keyword "domain_query".
To avoid breaking old code, sets the default value for "domain_query" as True.
EDIT: I forgot to mention: I have tested this with airnow_wrfchem.ipynb, adding False domain_query and giorgi and EPA regions.

Cheers
Pablo

### DIFF
--- a/docs/appendix/yaml.rst
+++ b/docs/appendix/yaml.rst
@@ -317,8 +317,8 @@ For example, ::
 **domain_type:** List of domain types to be plotted. These correspond with
 the columns in the observation file. (e.g., airnow: epa_region, state_name, 
 siteid, etc.). For automatic epa or giorgi boxes (if they are not included
-with the columns in the observation file), choose auto-region:epa or
-auto-region:giorgi. Take into account that auto-region:epa is only a rough
+with the columns in the observation file), choose ``auto-region:epa`` or
+``auto-region:giorgi``. Take into account that ``auto-region:epa`` is only a rough
 approximation, since it assumes perfect, rectangular lonlat boxes.
 
 **domain_name:** List of domain names to be plotted. If domain_type = all, all 
--- a/docs/appendix/yaml.rst
+++ b/docs/appendix/yaml.rst
@@ -316,7 +316,10 @@ For example, ::
 
 **domain_type:** List of domain types to be plotted. These correspond with
 the columns in the observation file. (e.g., airnow: epa_region, state_name, 
-siteid, etc.).
+siteid, etc.). For automatic epa or giorgi boxes (if they are not included
+with the columns in the observation file), choose auto-region:epa or
+auto-region:giorgi. Take into account that auto-region:epa is only a rough
+approximation, since it assumes perfect, rectangular lonlat boxes.
 
 **domain_name:** List of domain names to be plotted. If domain_type = all, all 
 data will be used and the domain_name is used only in the plot title. If 
--- a/docs/appendix/yaml.rst
+++ b/docs/appendix/yaml.rst
@@ -316,7 +316,8 @@ For example, ::
 
 **domain_type:** List of domain types to be plotted. These correspond with
 the columns in the observation file. (e.g., airnow: epa_region, state_name, 
-siteid, etc.). For automatic epa or giorgi boxes (if they are not included
+siteid, etc.).
+For automatic EPA or Giorgi region boxes (if they are not included
 with the columns in the observation file), choose ``auto-region:epa`` or
 ``auto-region:giorgi``. Take into account that ``auto-region:epa`` is only a rough
 approximation, since it assumes perfect, rectangular lonlat boxes.
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -1509,10 +1509,11 @@ class analysis:
 
                         # Query selected points if applicable
                         if domain_type != 'all':
-                            if "auto-region" in domain_type: 
-                                if 'epa_region' in domain_type:
+                            if domain_type.startswith("auto-region"):
+                                _, auto_region_id = domain_type.split(":")
+                                if auto_region_id == 'epa':
                                     bounds = get_epa_region_bounds(acronym=domain_name)
-                                elif 'giorgi_region' in domain_type:
+                                elif auto_region_id == 'giorgi':
                                     bounds = get_giorgi_region_bounds(acronym=domain_name)
                                 else:
                                     raise ValueError(
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -1522,10 +1522,12 @@ class analysis:
                                 elif domain_type == 'giorgi_region':
                                     bounds = get_giorgi_region_bounds(acronym=domain_name)
                                 else:
-                                    raise ValueError("Currently, region selections whithout a domain query have only"
-                                                     + "been implemented for Giorgi and EPA regions. You asked for"
-                                                     + f" {domain_type}. Soon, arbitrary rectangular boxes, US states and"
-                                                     + "others will be included.")
+                                    raise ValueError(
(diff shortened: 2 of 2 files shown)


melodies_monet/driver.py
```diff
@@ -1509,7 +1514,24 @@ def plotting(self):
...
+                            else: 
+                                if domain_type == 'epa_region':
+                                    bounds = get_epa_region_bounds(acronym=domain_name)
+                                elif domain_type == 'giorgi_region':
+                                    bounds = get_giorgi_region_bounds(acronym=domain_name)
+                                else:
+                                    raise ValueError("Currently, region selections whithout a domain query have only"
+                                                     + "been implemented for Giorgi and EPA regions. You asked for"
+                                                     + f" {domain_type}. Soon, arbitrary rectangular boxes, US states and"
+                                                     + "others will be included.")
```
zmoon, line 1528:
```suggestion
                                    raise ValueError(
                                        "Currently, region selections whithout a domain query have only "
                                        "been implemented for Giorgi and EPA regions. You asked for "
                                        f"{domain_type!r}. Soon, arbitrary rectangular boxes, US states and "
                                        "others will be included."
                                    )
```

melodies_monet/driver.py
```diff
@@ -1509,7 +1514,24 @@ def plotting(self):
...
+                                pairdf = pairdf_all.loc[
+                                             (pairdf_all["latitude"] > bounds[0])
```
zmoon, line 1530: Only works if region doesn't cross antimeridian and doesn't include N or S pole. And presumes bounds are in correct order and in [-180, 180) longitude format like the data. Fine for now but we might want to increase robustness in the future.

zmoon (commented):
I think the proposed config format (`domain_query=False`) is kind of confusing. In a sense it's still a query, just based on automatic lat/lon box. To me this feels more like a "domain query" than the original using site metadata already present in the dataframe.

Also, the Giorgi regions actually are rectangles so that's fine. But I would skeptical about suggesting people use the EPA region rectangles, which are pretty rough estimates.

I have [a little project](https://github.com/zmoon/epa-regions-python) for assigning to precise EPA regions. But it uses GeoPandas which is a heavy dependency.

rschwant:
@zmoon, should we call it something else then?

domain_metadata=True (default, use metadata from observation file to determine domain)
domain_metadata=False (use build in features in MONETIO to determine metadata since it is not provided?)

Open to other options too?

zmoon:
> should we call it something else then?

I feel it could be done just with `domain_type` with a special syntax. For example
- `domain_type: auto-region:giorgi` -- group on the fly using the rectangles (this PR method)
- `domain_type: giorgi_region` -- current behavior, using column `giorgi_region` in the dataframe

rschwant:
I like this method too. This may be easier to explain in the docs too.

melodies_monet/driver.py
blychs: Agreed. Since this is only implemented for Giorgi and EPA rectangles, it's not a problem for now. I suggest adding regionmask (check #272 ). This would, however, also require geopandas, since it's required by regionmask. I think that having it as an optional dependency would make sense.

blychs:
OK, I'll try to add that now.

melodies_monet/driver.py
blychs: Do you think we should add that check?

blychs:
> I think the proposed config format (`domain_query=False`) is kind of confusing. In a sense it's still a query, just based on automatic lat/lon box. To me this feels more like a "domain query" than the original using site metadata already present in the dataframe.
> 
> Also, the Giorgi regions actually are rectangles so that's fine. But I would skeptical about suggesting people use the EPA region rectangles, which are pretty rough estimates.
> 
> I have [a little project](https://github.com/zmoon/epa-regions-python) for assigning to precise EPA regions. But it uses GeoPandas which is a heavy dependency.

If we end up adding optional regionmask, that would include geopandas as a requirement.

melodies_monet/driver.py
```diff
@@ -1509,7 +1509,26 @@ def plotting(self):
 
                         # Query selected points if applicable
                         if domain_type != 'all':
-                            pairdf_all.query(domain_type + ' == ' + '"' + domain_name + '"', inplace=True)
+                            if "auto-region" in domain_type: 
+                                if 'epa_region' in domain_type:
+                                    bounds = get_epa_region_bounds(acronym=domain_name)
+                                elif 'giorgi_region' in domain_type:
```
zmoon, line 1515: Alternative
```suggestion
                            if domain_type.startswith("auto-region"):
                                _, auto_region_type = domain_type.split(":")
                                if auto_region_type == 'epa':
                                    bounds = get_epa_region_bounds(acronym=domain_name)
                                elif auto_region_type == 'giorgi':
```
Also, to me, since "auto-region" already has "region" in it, the identifier doesn't need to.

melodies_monet/driver.py
blychs: @rschwant @zmoon do we need anything else? I would like to have this solved before doing new PR's, to test the new code with this while avoiding branch conflicts.

zmoon:
@blychs do you want to mention these new option in the docs YAML section?

blychs:
I don't understand the failed tests...  I don't believe I changed any of that?

docs/appendix/yaml.rst
```diff
@@ -316,7 +316,10 @@ For example, ::
 
 **domain_type:** List of domain types to be plotted. These correspond with
 the columns in the observation file. (e.g., airnow: epa_region, state_name, 
-siteid, etc.).
+siteid, etc.). For automatic epa or giorgi boxes (if they are not included
+with the columns in the observation file), choose auto-region:epa or
+auto-region:giorgi. Take into account that auto-region:epa is only a rough
```
zmoon, line 321:
```suggestion
with the columns in the observation file), choose ``auto-region:epa`` or
``auto-region:giorgi``. Take into account that ``auto-region:epa`` is only a rough
```

blychs:
@zmoon One of the tests is failing because it does not find noaa   https://www.ncdc.noaa.gov/crn/  . Is it just out or does that link require updates?

rschwant:
We are linking to data that normally is stable, but is located at the Asheville office, so it's likely that the recent hurricane took down some of their servers. Let's see if it's back up and running tomorrow and go from there. We can re-run the checks then.

zmoon:
It seems the NCEI/NCDC websites are still having issues: https://www.ncei.noaa.gov/node/6696

rschwant:
@zmoon do you approve this now, since you have looked at this closer than me? If so, I will merge without the checks passing as it might take NOAA awhile to fix the servers in Asheville.

docs/appendix/yaml.rst
```diff
@@ -316,7 +316,10 @@ For example, ::
 
 **domain_type:** List of domain types to be plotted. These correspond with
 the columns in the observation file. (e.g., airnow: epa_region, state_name, 
-siteid, etc.).
+siteid, etc.). For automatic epa or giorgi boxes (if they are not included
```
zmoon, line 319:
```suggestion
siteid, etc.).
For automatic EPA or Giorgi region boxes (if they are not included
```

zmoon:
@blychs were you able to test that it seems to work?

blychs:
Wait, something is wrong. Nothing breaks (I had tested it) but it seems to not be doing anything. I'll fix it right away

blychs:
OK, it was in effect being skipped. I had tested it with the airnow_wrfchem example, nothing broke, and I was happy with it.
Now it should be fine. I just tested it with that example, running it with and without auto-region:epa and auto-region:giorgi.
I don't have an aircraft example though.
We were not supporting anything like this for satellite before. This might work for it but I haven't tested it.

PR #282, opened 2024-10-03
revise tropomi modules for monetio updates

mlirenzhenmayi:
update drivers, tropomi no2  - wrfchem paring


melodies_monet/driver.py
```diff
@@ -1756,7 +1759,8 @@ def plotting(self):
                             if filter_criteria and 'altitude' in filter_criteria:
                                 vmin_y2, vmax_y2 = filter_criteria['altitude']['value']
                             elif filter_criteria is None:
-                                if 'altitude' in pairdf.columns:
+                                #if 'altitude' in pairdf.columns: # pairdf is dataset object, don't have columns
+                                if 'altitude' in list(pairdf.data_vars.keys()):
```
blychs, line 1763: This can be just
```if 'altitude' in paridf.keys()```

I also did the same change to the hotfix

Comment (zmoon) [reply melodies_monet/driver.py:1763]: Also `if 'altitude' in pairdf` would work for both.

But for clarity I like
- `if var in df.columns` for pandas
- `if var in ds.variables` for xarray if the var might be a coordinate variable
  or `if var in ds.data_vars` if you know it's not a coord

File: melodies_monet/util/cal_mod_no2col.py [hunk 2]
```diff
@@ -28,23 +28,23 @@ def cal_model_no2columns(modobj):
...
+    # update, mli, to be coordinated with monetio
     no2    = modobj['no2']
-    ph     = modobj['PH']
-    phb    = modobj['PHB']
-    pb     = modobj['PB']
-    tdata  = modobj['T']
-    pdata  = modobj['P']
+    tb     = modobj['temperature_k']
+    pb2    = modobj['pres_pa_mid'] # time,z,y,x
+    zh     = modobj['zstag'] # time, bottom_up_stag, y,x
```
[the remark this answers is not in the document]
mlirenzhenmayi: revised to use dz_m directly in the new commit

melodies_monet/driver.py
mlirenzhenmayi: I revise it to if 'altitude' in pairdf.variables in the new commit, as the altitude might be a coordinate variable. I think this makes more sense.

mbruckner-work:
From what I can tell, the changes are fine for ensuring the TROPOMI NO2 capability works before the tutorial.

I would suggest adding a scaling to the paired observations and model data like was done for MOPITT CO (eg. linked codeblock). An argument to specify ylabel_plot (ylabel_plot: '$10^{18} molec/cm^{2}$' ) then should go in the yaml file with the variable name under the obs label (see control_mopitt.yaml). The axes on the Taylor plot will actually be readable this way.

https://github.com/NOAA-CSL/MELODIES-MONET/blob/c9a1ec1d29d273e7b4d04fd0478b56e5689b13e9/melodies_monet/util/satellite_utilities.py#L107-L112

blychs:
@mlirenzhenmayi , I would still go for keys(). This is already the solution in #284 , and .variables fails if it is DataFrame.
Otherwise, we would have to add checks to see whether it is a dataframe or a dataset.

zmoon:
@blychs for membership it's the same without `.keys()` (`if var in df_or_ds:`). Will it not always be an xarray dataset for satellite?

blychs:
If car in ds_or_df would work. Satellite is always ds, but that check is run for every case, including surface and aircraft, if I'm not wrong

melodies_monet/driver.py
```diff
@@ -1756,7 +1759,8 @@ def plotting(self):
                             if filter_criteria and 'altitude' in filter_criteria:
                                 vmin_y2, vmax_y2 = filter_criteria['altitude']['value']
                             elif filter_criteria is None:
-                                if 'altitude' in pairdf.columns:
+                                #if 'altitude' in pairdf.columns: # pairdf is dataset object, don't have columns
+                                if 'altitude' in pairdf.variables: # for xarray that might be a coordinate variable
```
zmoon, line 1763:
```suggestion
                                if 'altitude' in pairdf:
```
@blychs said https://github.com/NOAA-CSL/MELODIES-MONET/pull/282#issuecomment-2394967837

mlirenzhenmayi:
> From what I can tell, the changes are fine for ensuring the TROPOMI NO2 capability works before the tutorial.
> 
> I would suggest adding a scaling to the paired observations and model data like was done for MOPITT CO (eg. linked codeblock). An argument to specify ylabel_plot (ylabel_plot: '$10^{18} molec/cm^{2}$' ) then should go in the yaml file with the variable name under the obs label (see control_mopitt.yaml). The axes on the Taylor plot will actually be readable this way.
> 
> https://github.com/NOAA-CSL/MELODIES-MONET/blob/c9a1ec1d29d273e7b4d04fd0478b56e5689b13e9/melodies_monet/util/satellite_utilities.py#L107-L112

Thanks Maggie! Current tropomi modules can plot Taylor, although with messed units. But with the scaling applied, other plots like the map plot, box plot are messed up and need to be re-configured with units in the yaml file. I prefer to revise the Taylor plot module to make it more flexible to handle large units.

mbruckner-work:
> 
> Thanks Maggie! Current tropomi modules can plot Taylor, although with messed units. But with the scaling applied, other plots like the map plot, box plot are messed up and need to be re-configured with units in the yaml file. I prefer to revise the Taylor plot module to make it more flexible to handle large units.

I hadn't noticed an issue with maps, but generally am not using the box plot tool and haven't been checking it in my tests. I agree with @blychs that we can deal with solutions for this later.

mlirenzhenmayi:
> > > From what I can tell, the changes are fine for ensuring the TROPOMI NO2 capability works before the tutorial.
> > > I would suggest adding a scaling to the paired observations and model data like was done for MOPITT CO (eg. linked codeblock). An argument to specify ylabel_plot (ylabel_plot: '$10^{18} molec/cm^{2}$' ) then should go in the yaml file with the variable name under the obs label (see control_mopitt.yaml). The axes on the Taylor plot will actually be readable this way.
> > > https://github.com/NOAA-CSL/MELODIES-MONET/blob/c9a1ec1d29d273e7b4d04fd0478b56e5689b13e9/melodies_monet/util/satellite_utilities.py#L107-L112
> > 
> > 
> > Thanks Maggie! Current tropomi modules can plot Taylor, although with messed units. But with the scaling applied, other plots like the map plot, box plot are messed up and need to be re-configured with units in the yaml file. I prefer to revise the Taylor plot module to make it more flexible to handle large units.
> 
> Hi Meng and Maggie, I tried to do that in the plotting class and was not able to (for the TEMPO tool). I think that the problem is with the AxisArtist class from Matplotlib that the MONET uses, but I was not able to find an easy solution for that. In satplots_xr, I just added the option "normalize", which plots a normalized Taylor diagram... therefore having smaller numbers. I agree that revising the Taylor plot module is a better solution, and maybe someone else can find a way to do it (maybe there is a way with the GridHelperCurveLine?).
> 
> Having the option for an offset as suggested by Maggie is a better solution than mine, we could add that to the plt_grp in the yaml (and not the observations) for Taylor (even better, we could have both an offset and the option for normalizing it). However, I'd say that we can merge TROPOMI for now and deal with that after the deadline. Would that work? Cheers Pablo

Sounds good to me. - Meng

rschwant:
Is this ready now? Should we merge it in?

blychs:
@mbruckner-work , I believe we still need you to approve the review?

mbruckner-work (approved):

PR #284, opened 2024-10-04
Satellite plotting fix, OMPS pairing fix, and some additional time interval looping

mbruckner-work:
Fixes to satplots from @blychs.
OMPS-NM level 3 ozone pairing fix necessary for calculating daily mean from model's xarray dataset. 
Enables time looped processing for RAQMS model, MOPITT CO obs. Small syntax change to allow use of time looping and OMPS-NM level 2 from Suomi NPP, NOAA-20, and future OMPS-NM instruments.


blychs:
The code added by @mbruckner-work looks fine to me. Since this PR includes some of my own commits, I guess we want an independent reviewer though?
Could @rrbuchholz or @dwfncar do it?

dwfncar (approved):
All looks good.

melodies_monet/util/satellite_utilities.py
```diff
@@ -140,11 +140,12 @@ def omps_l3_daily_o3_pairing(model_data,obs_data,ozone_ppbv_varname):
     grid_adjust = xe.Regridder(model_data[['latitude','longitude']],obs_data[['latitude','longitude']],'bilinear')
     mod_col_obsgrid = grid_adjust(column)
     # Aggregate time-step to daily means
-    daily_mean = mod_col_obsgrid.groupby('time.date').mean(numeric_only=True).compute()
-    
+    #daily_mean = mod_col_obsgrid.groupby('time.date').mean().compute()
+    daily_mean = mod_col_obsgrid.resample(time='1D').mean()
```
zmoon, line 144: @blychs should there be `numeric_only=True` here?

zmoon: And @mbruckner-work just curious why the 91ad0c3111b8b042e9c92efb359b42bec41e15e5 change? Shouldn't `.groupby('time.date').mean()` and `.resample(time='1D').mean()` give the same result?

mbruckner-work: @zmoon The data is in xarray here. Putting in 'numeric_only=True' causes the code to fail.

blychs: I believe that @mbruckner-work found that it didn't work, since it is a Dataset instead of a DataFrame, and needed to be taken out.

mbruckner-work: The https://github.com/NOAA-CSL/MELODIES-MONET/commit/91ad0c3111b8b042e9c92efb359b42bec41e15e5 change was done because of the time format of the paired dataset. '.groupby('time.date').mean()' changes the time format to Timestamp while '.resample(time='1D').mean()' keeps the format as datetime64. 
The old way resulted in an error with the time filtering in the plotting code in my most recent tests, hence the change.

zmoon: Thanks @mbruckner-work . Just wanted to make sure we are keeping up with @blychs efforts to supports pandas 2. You could remove the commented out lines that (no longer?) work.

rschwant:
@mbruckner-work and @blychs, there is a small conflict now. Will Meng's method work here for you all too?

blychs:
@rschwant It should work (and probably be a better solution than ours).

rschwant:
I'm testing that Meng's update still works with the aircraft code now. Do you all want to test that it works for your satellite options? Or if you are pretty confident it will work. I can choose Meng's option. We can merge it in. And everyone can retest everything this afternoon.

blychs:
I merged Meng's version, my code works fine and I don't think this can break Maggie's, but do you want to test it, just in case, @mbruckner-work ?

mbruckner-work:
My test cases worked fine.

blychs:
Then I think it could be merged, @rschwant , unless we need another review.

rschwant:
I'm worried that both of your all's fix for the filter_criteria that Quazi set up breaks the altitude feature Quazi added @quaz115. I'm looking into it.

blychs:
That's surprising. Is it breaking in the same line? Because we would expect both our previous solution and Meng's solution to not affect that, since the
```
if 'altitude' in ds_or_df:
```
should work worth for df and ds.
We could modify it to 
```
if 'altitude' in ds_or_df.keys():
```

or, worst case scenario, for a dirty solution
```
if isinstance(pairdf, pd.DataFrame) and 'altitude' in pairdf:
```

melodies_monet/util/satellite_utilities.py
zmoon: It would be great to remove those lines before merging this.

zmoon: Thanks @mbruckner-work appreciate it

rschwant:
Okay it's fine. I'm having trouble calling the logic that Quazi set up as I don't fully understand the yaml options, but we update this later. The example runs well, so this can be pulled in.

PR #289, opened 2024-10-07
Keep ISH tests from failing CI if data unreachable

zmoon:
Let's get that ✔ back.

xref: https://github.com/noaa-oar-arl/monetio/pull/202

### DIFF
--- a/melodies_monet/tests/test_get_data_cli.py
+++ b/melodies_monet/tests/test_get_data_cli.py
@@ -7,9 +7,25 @@ Check for consistency with the tutorial datasets and that options work.
 import subprocess
 
 import numpy as np
+import pytest
+import requests
 import xarray as xr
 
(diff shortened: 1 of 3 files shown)


melodies_monet/tests/test_get_data_cli.py
```diff
@@ -7,10 +7,26 @@
...
 
+try:
+    requests.head("https://www1.ncdc.noaa.gov/pub/data/noaa/")
+except Exception:
+    ish_reachable = False
+else:
+    ish_reachable = True
+
+try:
+    r = requests.head("https://www1.ncdc.noaa.gov/pub/data/noaa/isd-lite/")
```
zmoon, line 24: Don't need to store.
```suggestion
    requests.head("https://www1.ncdc.noaa.gov/pub/data/noaa/isd-lite/")
```

PR #291, opened 2024-10-08
Run the two low-mem notebooks in docs build

zmoon:
* added numba to the docs env and adjusted some utils so `xesmf` isn't imported at module level, allowing one to still build the docs without `xesmf` (e.g. on Windows)
* notebooks where all cells have output should be automatically skipped anyway, but some don't seem to meet that, so I added ignores for all of those


rschwant (approved):
Looks great!

rrbuchholz (approved):
It looks fine to me as well.

zmoon:
Thanks!

PR #296, opened 2024-10-10
CLI: Include exception type name in non-debug message

zmoon:
With this change, there's a little bit more info about the error in `--no-debug` mode (default)

* [x] add a simple test


rschwant (approved):
Looks good! I like the additional tests.

PR #297, opened 2024-10-10
CLI: address xarray `.dims` -> `.sizes` warning

zmoon:
`ds.dims` will be changed to tuple of dim names in the future (like [`da.dims`](https://docs.xarray.dev/en/stable/generated/xarray.DataArray.dims.html) currently returns).


rschwant (approved):
Looks good!

PR #299, opened 2024-10-22
Update docs env

zmoon:
We could also update the docs to recommend Python ~~3.10~~ 3.11 instead of 3.9? @rschwant


rschwant (approved):
Looks good to update the environment. And I'm happy to move to later versions of Python because we can test the code with this now too.

rschwant (approved):
Looks Great!

PR #304, opened 2024-10-29
longitude wrap satellite level 3 pairing

mbruckner-work:
Add periodic=True for regridding model data to level 3 satellite data to allow regrid function to handle longitude wrap. 

This is a temporary work around as it assumes the model is a global model. To properly allow for flexible comparison of satellite level 3 products and any model domain (regional or global), need a function to subset satellite observations to the model domain and/or associate a model domain type flag with the model object.

### DIFF
--- a/melodies_monet/util/satellite_utilities.py
+++ b/melodies_monet/util/satellite_utilities.py
@@ -78,7 +78,7 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname):
         raise
     # initialize regridder for horizontal interpolation 
     # from model grid to MOPITT grid
-    grid_adjust = xe.Regridder(model_obstime[['latitude','longitude']],obs_data[['lat','lon']],'bilinear')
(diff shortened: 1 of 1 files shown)


blychs:
If I'm not wrong, reading the docs, it shouldn't do anything if the model isn't global, should it?
In that case, it wouldn't be detrimental, and I can't find any disadvantage to your approach.

Otherwise, a different workaround would be defining a function which, just in pseudocode, does the check :
def isglobal(modobj: xr.Dataset) -> Boolean:
     return test_isglobal
grid_adjust = xe.Regridder(model_obstime[['latitude','longitude']],obs_data[['lat','lon']],'bilinear',periodic=test_isglobal)

blychs (approved):
Everything looks right to me. 
According to the docs, I don't think that the periodic=True parameter should be a problem, but we might still need to test it.

blychs:
If this gets merged, we could add an issue to add it to the YAML file.

dwfncar (approved):

PR #309, opened 2024-11-04
Add option to disable adding logo to plots

### LINKED ISSUE #308: Make logo optional
We need to make the logo optional so people can choose to turn this off especially for publication plots. All examples need to default to this on though.

zmoon:
```yaml
analysis:
  add_logo: false
```

Resolves #308

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -738,6 +738,8 @@ class analysis:
         self.obs_gridded_data = {}
         self.obs_gridded_count = {}
         self.obs_gridded_dataset = None
+        self.add_logo = True
+        """bool, default=True : Add the MELODIES MONET logo to the output plots."""
 
     def __repr__(self):
         return (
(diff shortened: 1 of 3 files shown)


zmoon:
I tested with idealized example and it worked, but maybe someone could test a more complicated example just to make sure.

melodies_monet/driver.py
```diff
@@ -738,6 +738,8 @@ def __init__(self):
         self.obs_gridded_data = {}
         self.obs_gridded_count = {}
         self.obs_gridded_dataset = None
+        self.add_logo = True
+        """bool, default=True : Add the MELODIES MONET logo to the output plots."""
```
zmoon, line 742:
```suggestion
        """bool, default=True : Add the MELODIES MONET logo to the plots."""
```

mbruckner-work:
Tested with a satellite test case. The savefig calls within some of the satellite/xarray plotting routines currently include `decorate = True` as a hardwired argument. 
When these are removed, `add_logo` works as intended.

zmoon:
> Tested with a satellite test case. The savefig calls within some of the satellite/xarray plotting routines currently include `decorate = True` as a hardwired argument. When these are removed, `add_logo` works as intended.

Thanks @mbruckner-work for pointing that out, I will address.

mbruckner-work (approved):
Looks good! 
From my review of the driver and the plotting routine sets, satplots was the only one with hardwired `decorate`.

PR #312, opened 2024-11-14
Fix for np.float deprecation

mbruckner-work:
Pairing code for TROPOMI NO2 used `np.float` in one line. 
The np.float alias was deprecated in the Numpy 1.20.0 release (see https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations).


blychs:
This looks to me as ready to be merged.

Issue #307, opened 2024-11-01
Ground and model with same name (csv).


blychs:
Hi all,
There is a bug in surfplots if ground stations and model have the same variable named. The bug is caused because monet changes the name by adding "new", whereas the pairing tool in melodies-monet tool.py utility (mobile_and_ground_pair) uses directly pandas merge_asof, which in that case changes the name by adding _x, _y, etc.
Then, the plotting tool looks for the variable "_new", which does not exist.
Probably the easiest fix is adding _new to repeated variables, and sticking to the MONET method, although I tend to think that the pandas way of dealing with it is probably better.
What do you think, @zmoon ?
Cheers
Pablo

zmoon:
@blychs this could likely be done in your pandas merge with `suffixes=(None, '_new')` (which is how monet combinetool does it too). Depending on context of course, I do like the approach of only adding a suffix to the "new" column.

blychs:
Solved in #313

PR #313, opened 2024-11-19
Bugfix: suffix when merging mobile_or_aircraft

blychs:
Previously it would add the defaul _x, _y, but MELODIES-MONET expects '', '_new'. This addresses #307


blychs:
I'm trying to understand the checks that failed, since this PR is only a very minor change?

zmoon:
The ISH downloading has been flaky the last few days. Also the docs link check found one broken.

zmoon (approved):

PR #316, opened 2024-11-21
Changes to MOPITT pairing

mbruckner-work:
Adds utility for sampling model data to satellite local overpass time. 

Modifies MOPITT L3 pairing to calculate model averages using only model data at the satellite overpass time. 
Also ensures that model is paired to correct obs time. Previously assumed that model and obs datasets were on same time at that point, but this isn't true if there are gaps in the obs dataset.

New argument for the yaml file if model output hasn't been pre-processed to monthly mean: mod_to_overpass. 
 

Note: Insert/delete flagging includes some commits from other recent pull requests (#304, #312 ).


melodies_monet/driver.py
```diff
@@ -287,6 +287,13 @@ def open_sat_obs(self, time_interval=None, control_dict=None):
                 else: flst = self.file
                 self.obj = mio.sat._mopitt_l3_mm.open_dataset(flst, ['column','pressure_surf','apriori_col',
                                                                           'apriori_surf','apriori_prof','ak_col'])
+
+                # Determine if monthly or daily product and set as attribute
+                if 'MOP03JM' in glob(self.file)[0]:
```
rrbuchholz, line 292: Add in options for MOP03NM and MOP03TM for NIR and TIR only L3 files

melodies_monet/driver.py
```diff
@@ -1309,6 +1321,11 @@ def pair_data(self, time_interval=None):
                 # TODO: add other network types / data types where (ie flight, satellite etc)
                 # if sat_swath_clm (satellite l2 column products)
                 elif obs.obs_type.lower() == 'sat_swath_clm':
+                    # grab kwargs for pairing. Use default if not specified
+                    if obs.obs_type.lower() in list(self.pairing_kwargs.keys()):
+                        pairing_kws = self.pairing_kwargs[obs.obs_type.lower()]
+                    else: 
+                        pairing_kws = {'apply_ak': True, 'mod_to_overpass': False}
```
rrbuchholz, line 1328: Need to add an error message as to why MM might break if no AK is available.

rrbuchholz (approved):
Looks good to me. Ready to merge once those other 2 Level3 options are added.

rrbuchholz:
Looks like Maggie made those few updates and it is ready to merge -- @rschwant or @zmoon do you agree?

rrbuchholz (approved):
Looks good.

melodies_monet/util/satellite_utilities.py
```diff
@@ -32,6 +32,56 @@ def vertical_regrid(input_press, input_values, output_press):
...
+    local_utc_offset = np.zeros([ny,nx],dtype='timedelta64[ns]')
+    # pandas timedelta calculation doesn't work on ndarrays
+    for xi in np.arange(nx):
+        local_utc_offset[:,xi] = pd.to_timedelta((modobj['longitude'].isel(x=xi)/15).astype(np.int64),unit='h')
+    
+    # initialize local time as variable
+    modobj['localtime'] = (['time','y','x'],np.zeros([nmt,ny,nx],dtype='datetime64[ns]'))
+    # fill
+    for ti in np.arange(nmt):
+        modobj['localtime'][ti] = modobj['time'][ti].data + local_utc_offset
```
zmoon, line 65: You may want to round `lon/15` instead of truncating by converting to int. You could try the approach here to get around the `pd.to_timedelta` limitation:
```python
local_utc_offset = (modobj['longitude'] / 15).round().astype("timedelta64[h]")
```
Then you should be able to do
```python
modobj['localtime'] = modobj['time'] + local_utc_offset
```

melodies_monet/driver.py
zmoon: Seems like the message could be made a bit more helpful by telling you *why* it's being set to true (defaulting since not set) and how you (the user) could fix it.

melodies_monet/driver.py
```diff
@@ -1326,6 +1339,12 @@ def pair_data(self, time_interval=None):
                 # TODO: add other network types / data types where (ie flight, satellite etc)
                 # if sat_swath_clm (satellite l2 column products)
                 elif obs.obs_type.lower() == 'sat_swath_clm':
+                    # grab kwargs for pairing. Use default if not specified
+                    if obs.obs_type.lower() in list(self.pairing_kwargs.keys()):
+                        pairing_kws = self.pairing_kwargs[obs.obs_type.lower()]
+                    else: 
+                        print('WARNING: The satellite pairing option apply_ak is being set to True. Pairing will fail if there is no AK available.')
+                        pairing_kws = {'apply_ak': True, 'mod_to_overpass': False}
```
zmoon, line 1347: Currently you'd have an issue later if `pairing_kwargs` for sat were set in the control but didn't contain `apply_ak`, e.g. just `mod_to_overpass` or another unrelated setting. An alternative approach would be to set `pairing_kws` to your desired defaults first and then update it based on `pairing_kwargs` from the control. That way you know everything is set. There would still be an opportunity to let the user know that they may need to set `apply_ak`.

zmoon (commented):
Added a few comments.

melodies_monet/util/satellite_utilities.py
```diff
@@ -89,15 +154,18 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname):
...
+                                       obs_data['pressure'][t].values)
     
     # apply AK
     ## log apriori and model data
     log_ap = np.log10(obs_data['apriori_prof'])
     log_mod = np.log10(co_regrid)
     diff_arr = log_mod-log_ap
     ## smooth/apply ak
-    smoothed = obs_data['apriori_col'] + (obs_data['ak_col']*diff_arr).sum(dim='alt')
+    smoothed = obs_data['apriori_col'] + (obs_data['ak_col']*diff_arr).sum(dim='alt', skipna=False)
```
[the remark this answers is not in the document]
mbruckner-work: Yes, this is something we needed for the regional model pairing. MOPITT L3 is on a global 1 $^\circ$ x 1 $^\circ$ grid and the apriroi_col will be defined outside of a regional model domain.

In the future we should probably add a limiter on the obs data that removes anything outside the model domain.

docs/appendix/yaml.rst
```diff
@@ -109,9 +115,7 @@ data (e.g., surf_only: True).
 Typically this is set at the horizontal resolution of your model * 1.5. Setting 
 this to a smaller value will speed up the pairing process. 
 
-**apply_ak:** This is an optional argument used for pairing of satellite data. This
-should be set to True when application of satellite averaging kernels or apriori data 
-to model observations is desired. 
+**apply_ak:** Deprecated and removed. Instead, specify ``pairing_kwargs`` in the analysis section.
```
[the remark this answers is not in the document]
mbruckner-work: done.

examples/yaml/control_ufsaqm_mopitt.yaml
```diff
@@ -0,0 +1,79 @@
...
+        debug: False
+        pairing_kwargs:
+                sat_grid_clm:
+                        apply_ak: True
+                        mod_to_overpass: True
+model:
+        ufsaqm: 
+                files: '/scratch1/BMC/rcm2/rhs/monet_example/AEROMMA/UFS-AQM/cmaq54_OriRave1/aqm.202306*/12/aqm.t12z.dyn.*.nc'
+                mod_type: 'rrfs'
+                is_global: False
```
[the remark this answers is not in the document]
mbruckner-work: Fixed and added a description to YAML doc.

melodies_monet/driver.py
zmoon: I think this is resolved now.

zmoon (approved):

Issue #321, opened 2024-11-27
Update headers in src files


rschwant:
We need to update the headers in all the src files to remove this line:
"# Copyright (C) 2022 National Center for Atmospheric Research and National Oceanic and Atmospheric Administration"

Update the checks to only check for this line on top of all source code files instead of both lines
"# SPDX-License-Identifier: Apache-2.0"

And then add a NOTICE file and copy the copyright above there instead. Then we don't have separate copyright dates for new pieces of code and we do not have to update these headers every year to add in 2022 - 2024 etc.

Since we are just moving this notice to one location, no copyright changes to the code occur. This is just restructuring and aligns us with how the apache license recommends you set up your repository here: https://infra.apache.org/licensing-howto.html

Let's do this in the lull after we close out our current PRs after conferences and before winter break since this will result in changes to a lot of files and I don't want to cause merge conflicts. @blychs can you merge your TEMPO updates in without updating this for now and we will fix this after AGU? Or do we need to fix this before your PR goes through?

rschwant:
This is updated with this PR:
https://github.com/NOAA-CSL/MELODIES-MONET/pull/332

PR #326, opened 2024-12-31
Feature/region support

blychs:
Adds new region support in the form of a specific utility (util/region_select.py). The docs are updated accordingly.
This includes:
Old capabilities, but using xarray instead of pandas (and the `.where` method instead of a query). The query is still and option in the utility, but due to how I change the driver, I don't see how it would really be required by the code. xarray has a "query" method, but I couldn't make it work and I believe it requires the use of a dimension to query along, which I don't think we want.

New, advanced region support.
`auto-region` now includes `auto-region:custom`, allowing the user to provide a lonlat box. It is currently somewhat limited, though, and it cannot cross the antimeridian. The box has to be provided in the new keyword `domain_info`.

New, advanced region support with `regionmask`. These require the use of the new keyword `domain_info`. The new capabilities include:
 - Defining one or more polygons with `custom:auto_polygon` in `domain_type`. Holes in the polygon are permitted by regionmask, but I have not added that capability yet (I am not sure about the best way to do it, since it would require turning things into another dictionary, I believe).
 - Using `regionmask`'s `defined_regions` method with `custom:defined_regions`. These need to be defined, once again, in `domain_info`.
 - Using a shapefile/geojson with `custom:custom_file`. The path or URL of the file need to be defined in `domain_info`. There are a few undesirable thing about how I did the automatic download, though, and if you have better suggestions I'd be happy to change the code:
     - I am not using `pooch`, but downloading the files manually. The reason for this is that I was not able to find a way to tell `pooch` to use the content-disposition keyword of a URL when downloading, leading to errors when the URL did not end with the name of the file. I'd rather not make the user provide that, since it's not always certain. _I'm sure that there must be a way, but I could not find it_.
     - The code does not check if the file exists, and just overwrites it. If you ask for multiple tasks using the same domain, it downloads it again every time. This is silly, and could be avoided by using `pooch`, which would also test the checksum. Once again, I could not find an easy way to avoid this when the URL does not end with the appropriate file name, which happens quite often (for example, in my tests).

If you have any solutions for this, I would appreciate them. Otherwise, I'd suggest moving forward with this.
Please check also the changes in the docs. My English is far from perfect and some proofreading (and corrections!) by people with a better English than me would be great.

I am uploading here the yaml file I used to test all of this, so that it can get tested again. I tested it against surface data. Although I didn't do a complete test against TEMPO (I didn't want to add changes to the driver until I fix the merge conflicts, hopefully tomorrow) I did test the individual functions and plotted the results. Testing against AEROMMA data for aircraft would be great. Otherwise, I can try it with ASIA-AQ. Since GitHub does not allow me 
I also provide here my functions for testing individual options. I did not build proper unit tests, since just asserting whether the types of data are correct or not completely NaN does not seem as useful as looking at the plots.

You will see that I did not plot the multiboxplot nor the scorecards with the regions defined with regionmask. There is probably a way to do it (by adding the mask to the dataset and not only selecting the region, possibly in a copy to avoid changing it), but it seemed a little bit confusing.
[test_regionmask_tool.zip](https://github.com/user-attachments/files/18279502/test_regionmask_tool.zip)

Cheers,
Pablo

Edit: I requested some reviewers that seemed reasonable, but feel free to change that.


blychs:
The test that is failing is the Copyright notice, which I wrote according to #321

docs/appendix/yaml.rst
```diff
@@ -321,6 +321,22 @@ For automatic EPA or Giorgi region boxes (if they are not included
...
+
+If you have ``regionmask`` installed, you can also use it for advanced region support.
+These regions can be arbitrary, and its use require providing ``domain_type`` parameters starting
+with ``custom:``.
+There are three ways to use ``regionmask``. ``custom:auto_polygon`` lets the user define their own
+polygon in the section ``domain_info``, using the keyword ``mask_info``.
+``custom:defined_region`` lets the user utilize any region predefined by 
+`regionmask <https://regionmask.readthedocs.io/en/stable/>`__, defined in ``domain_info`` using
+the keywords ``name_regiontype`` and ``region``.
+The third option is using the keyword `custom:custom_file`, which is defined in ``domain_info`` with
```
zmoon, line 336: What about these names instead?:
* `custom:polygon`
* `custom:defined-region` or `custom:natural-earth` (or could this be combined with the `file` option?)
* `custom:file`

docs/appendix/yaml.rst
```diff
@@ -321,6 +321,22 @@ For automatic EPA or Giorgi region boxes (if they are not included
 with the columns in the observation file), choose ``auto-region:epa`` or
 ``auto-region:giorgi``. Take into account that ``auto-region:epa`` is only a rough
 approximation, since it assumes perfect, rectangular lonlat boxes.
+If you only need a rectangular, lonlat box which does not cross the antimeridian, you can use
+``auto-region:custom_box``, which needs to be combined with the ``domain_info`` parameter and
```
zmoon, line 325: Could this one be `custom:box` instead?

docs/appendix/yaml.rst
blychs: custom:defined-region looks good. I am not sure of all of the defined regions are in NaturalEarthData, so just in case I'd rather keep that one.
custom:file seems good to me. I think it might be more clear to keep it separate from defined-region since that requires the user to browse a bit regionmask, whereas the file option should only require the file.

docs/appendix/yaml.rst
blychs: Hi zmoon. I was thinking, shouldn't it be `auto-region:box` rather than `custom:box`? I was reserving the custom: `domain_type` for domains that require `regionmask`

blychs: I changed it to auto-region:box. Closing

blychs:
Hi all, I would like to leverage on this PR for xarray_plots. Therefore, the sooner we can get this merged, the better. Let me know any edits that should be done.
I would like to finalize xarray_plots as soon as possible. I've been checking and think that I might be able to use this to get rid of pandas for surface and satplots. I haven't looked at aircraftplots yet.

docs/appendix/yaml.rst
zmoon: Sorry for not responding earlier. My thought was that `custom:box` made more sense since it's a custom box that you have to specify, it's not fully automatic like the auto EPA region estimate option, for example.

blychs: I get the point, and it does make sense. I was keeping the `custom:` label for tools using regionmask, and the `auto-region:` for those that do not, and choosing which type of function to use based on that. That was also used in the code workflow, which is using that to choose what to use. I could modify it the way you suggest by adding a special case (basically, an `or` in the `if` condition that chooses the type of region), although it would make the code slightly worse, if you think that the user experience would be better.

melodies_monet/driver.py
```diff
@@ -1533,9 +1536,15 @@ def plotting(self):
...
+                            p_region = select_region(p.obj, domain_type, domain_name, domain_info)
+                        else:
+                            p_region = p.obj
+
+                        
+                        if obs_type in ["sat_swath_sfc", "sat_swath_clm", "sat_grid_sfc",
+                                        "sat_grid_clm", "sat_swath_prof"]:
                              # convert index to time; setup for sat_swath_clm
                             
                             if 'time' not in p.obj.dims and obs_type == 'sat_swath_clm':
```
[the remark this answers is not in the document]
rschwant: Yes, line 1541 (if statement) and 1543 (and swap). I guess the if statement wouldn't be different, but best to update for consistency.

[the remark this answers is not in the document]
blychs: Done.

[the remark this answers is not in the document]
rschwant: It is clearer if box is in the custom option instead of the auto-region option, but I do see why you separated it with regionmask. If it's not too complicated to move it, this is probably the clearest way forward, since most users will not understand the nuance of custom using the regionmask package.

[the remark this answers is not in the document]
blychs: Done

docs/appendix/yaml.rst
blychs: Changes done to custom:box.

docs/users_guide/region_selection.rst
```diff
@@ -0,0 +1,54 @@
+Selecting regions
+=================
+
+MELODIES-MONET has different ways to mask/select a specific region.
+
+The easiest, and often most precise, is to have that region defined in the observations/paired data.
+In this case, the variable/column that defines the region of interest should be selected as 
+`domain_type` in the YAML file, and the name of the region to select should be provided as `domain_name`.
+
+If that data are not available in the observations, another option is to provide a lonlat box, which
```
rschwant, line 10:
```suggestion
If the regional metadata are not available in the observations, another option is to provide a lonlat box, which
```

docs/users_guide/region_selection.rst
```diff
@@ -0,0 +1,54 @@
...
+can be defined by setting `domain_type: auto-region:xxxxx`, where `xxxxx` can be `epa`, `giorgi` or
```
rschwant, line 11:
```suggestion
can be defined by setting `domain_type: auto-region:xxxxx`, where `xxxxx` can be `epa` or `giorgi`, or setting
```

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+
+    Returns
+    -------
+    xr.Dataset | pd.DataFrame
+        Region selected. The type will be the same as the input data.
+    """
+
+    if domain_type == "all":
+        return data
+    if domain_type.startswith("auto-region") or (domain_type == "custom-box"):
```
rschwant, line 305: Should custom-box here be custom:box?

rschwant (commented):
Just a couple small updates, I think and this is ready.

melodies_monet/util/region_select.py
blychs: Yes, it was a typo. Fixed

rschwant (approved):
Looks great, thanks for making these updates!

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+
+try:
+    import geopandas as gp
+    from shapely.geometry import MultiPolygon, Polygon
+except ImportError:
+    gp = None
+try:
+    import regionmask
+except ImportError:
+    regionmask = None
```
zmoon, line 22: Currently you're only later checking if regionmask is `None` or not. The rest of these will produce errors that could be confusing if the things are not installed. You'd be better off doing the imports where you need them in the function bodies.

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+        bounds = get_epa_region_bounds(acronym=domain_name)
+    elif auto_region_id == "giorgi":
+        bounds = get_giorgi_region_bounds(acronym=domain_name)
+    elif auto_region_id == "box":
+        bounds = domain_info["bounds"]
+    else:
+        raise ValueError(
+            "Currently, auto-region selections whithout a domain query have only "
+            "been implemented for Giorgi and EPA regions. You asked for "
+            f"{domain_type!r}. If you need more capabilities, checkout the custom:"
```
zmoon, line 258:
```suggestion
            f"{domain_type!r}. If you need more capabilities, check out the custom: "
```
I would think there should be a space after "custom:", but not sure.

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+    url : str
+        URL of file to download
+    verbose : boolean
+        Whether to add verbosity
+    Returns
+    -------
+    str
+       Path to downloaded file
+    """
+    r = requests.get(url, allow_redirects=True, timeout=10)
```
zmoon, line 39:
```suggestion
    r = requests.get(url, allow_redirects=True, timeout=10)
    r.raise_for_status()
```

melodies_monet/util/region_select.py
blychs: I was doing it this way to stick to PEP8:
" Imports are always put at the top of the file, just after any module comments and docstrings, and before module globals and constants" . I have found that other codebases use this `None` trick and was copying that idea. If you think it's worth changing anyway, I'm happy to do so.

melodies_monet/util/region_select.py
blychs: Yes, I think you are right

melodies_monet/util/region_select.py
blychs: Seems like a good idea

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+    fname = re.findall("filename=(.+)", r.headers.get("content-disposition"))[0]
+    if len(fname) == 0:
+        fname = url.rsplit("/", 1)[1]
+    if fname[0] == '"':
+        fname = fname[1:-1]
```
zmoon, line 45: As it was, if Content-Disposition was missing the `re.findall` will error. And if it was there but `filename=` wasn't in it, `[0]` would error. And if `filename` wasn't last, the match would've included other directives. This should be a bit more robust:
```suggestion
    fname = None
    content_disposition = r.headers.get("Content-Disposition")
    if content_disposition:
        m = re.search(r"filename=(.+)[;$]", content_disposition)
        if m is not None:
            fname = m.group(1).strip('"')
    if fname is None:
        fname = url.rsplit("/", 1)[-1]
```

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,321 @@
...
+        return data
+    if domain_type.startswith("auto-region") or (domain_type == "custom:box"):
+        data_masked = create_autoregion(data, domain_type, domain_name, domain_info)
+    elif domain_type.startswith("custom"):
+        if regionmask is None:
+            raise ImportError(
+                "regionmask is not installed, try either auto-region: options or custom:box."
+            )
+        if domain_info is None:
+            raise KeyError("If regionmask is used, domain_info must exist.")
```
zmoon, line 314: Maybe it makes sense to move these to `control_custom_mask()`

melodies_monet/util/region_select.py
zmoon: >  Imports are always put at the top of the file

That's good and fine for std lib and required dependencies. And if the import time isn't an issue.

And this `None` trick is fine (or using another variable to keep track of if you have the import available or not) if it's just one or two imports you need, but it can get unwieldy.

As it is, you only check for regionmask in one place. Those other functions _could be_ called independently and the errors will be confusing if regionmask/Shapely/GeoPandas are not installed. You should at prefix the names of those functions with `_` to indicate you don't intend for them to be called directly.

> If regionmask is installed, then geopandas and shapely area as well. 

They _should be_, but it's not guaranteed. For example, if we were writing an optional dependency list, we would include all three of them, not just regionmask, since we are using all three explicitly here.

melodies_monet/util/region_select.py
zmoon: Otherwise there's no error if your URL is bad or something wrong with the server, etc.

melodies_monet/util/region_select.py
DWesl: Would `filename=([^;]+)(?=;|$)"` work better for the filename regex? [`$` in a character class will match only a $ in the input](https://docs.python.org/3/howto/regex.html#more-metacharacters) while this looks like it's trying to match end-of-line, and [`(?=` starts a lookahead assertion](https://docs.python.org/3/howto/regex.html#lookahead-assertions) (as an alternative, [`(?:` starts a non-capturing group](https://docs.python.org/3/howto/regex.html#non-capturing-and-named-groups) that won't appear in `m.groups()`)  I don't see anything there to keep the `.+` from gobbling the next item: [the non-greedy `.+?` should also work](https://docs.python.org/3/howto/regex.html#greedy-versus-non-greedy)

Alternately, `[part.split("=")[1] for part in content_disposition.split(";") if "filename=" in part]`

[This page](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) suggests the server might provide `filname*` in addition to or instead of `filename`, though that value would require additional processing.

It doesn't look like pooch has a way to set the file name after the download has started, but you might be able to guess from the URL.   Unfortunately, there's enough variation in how the URL encodes information about the resource it points to that a universal method of composing a simple filename obviously related to the URL is probably not possible.

Pooch's default involves a hash of the URL, would using just the hash without the guess at the filename work?  It's not as obviously related to the URL if you want to find the downloaded file in pooch's cache, but the code will be able to reconstruct it easily.

zmoon: Thanks @DWesl for the help and explanations. Turns out I hadn't properly checked the end-of-line case. Updated to the non-greedy with (non-capturing) group for the end.

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+
+
+def _download_with_name(url, verbose=False):
+    """Automates download while reading content disposition if needed.
+
+    Parameters:
+    -----------
+    url : str
+        URL of file to download
+    verbose : boolean
```
zmoon, line 32:
```suggestion
    verbose : bool
```

melodies_monet/util/region_select.py
```diff
@@ -0,0 +1,320 @@
...
+    Returns
+    -------
+    xr.Dataset
+        mask for the input data
+    """
+
+    if mask_url is not None and mask_path is not None:
+        warnings.warn(
+            "mask_url and mask_path provided. Only one can be used."
+            + "Selecting mask_path and discarding URL."
```
zmoon, line 157:
```suggestion
            "mask_url and mask_path provided. Only one can be used. "
            "Selecting mask_path and discarding URL."
```

melodies_monet/util/region_select.py
blychs: I'm not very good with regex. I just committed the last version from @zmoon, does it need to match @DWesl ? I can update it if required.

zmoon: @blychs the one you committed should be fine, that was based on @DWesl comments.

zmoon:
@blychs did you have any thoughts about these pooch comments [from @DWesl](https://github.com/NOAA-CSL/MELODIES-MONET/pull/326#discussion_r1953478536)?

> It doesn't look like pooch has a way to set the file name after the download has started, but you might be able to guess from the URL. Unfortunately, there's enough variation in how the URL encodes information about the resource it points to that a universal method of composing a simple filename obviously related to the URL is probably not possible.
>
> Pooch's default involves a hash of the URL, would using just the hash without the guess at the filename work? It's not as obviously related to the URL if you want to find the downloaded file in pooch's cache, but the code will be able to reconstruct it easily. It might be worth adding some part of the domain/hostname in the hash to get a little association between URL and cached file name and to reduce the chance of filename collisions.

melodies_monet/util/region_select.py
zmoon: Going to resolve this and move the pooch discussion to main thread.

melodies_monet/util/region_select.py
blychs: Done. Doing that also allows for control_custom_mask to be public again.

melodies_monet/util/region_select.py
blychs: > It doesn't look like pooch has a way to set the file name after the download has started, but you might be able to guess from the URL. Unfortunately, there's enough variation in how the URL encodes information about the resource it points to that a universal method of composing a simple filename obviously related to the URL is probably not possible.

This does not always work. E. g., if you look at the example in the docs that I added, the state of Colorado has URL's for their shapefiles that don't show the name at all. It is more than good enough in _most_ cases, but that few that don't work are, IMHO, bad enough that it's not the best for this generalized tool.


> Pooch's default involves a hash of the URL, would using just the hash without the guess at the filename work? It's not as obviously related to the URL if you want to find the downloaded file in pooch's cache, but the code will be able to reconstruct it easily. It might be worth adding some part of the domain/hostname in the hash to get a little association between URL and cached file name and to reduce the chance of filename collisions.

I am not sure if that would work, though. Regionmask assumes the format from the extension if I'm not wrong, and I'm definitely doing that for zipped files. That being the case, if I don't have a name containing the extension, this would fail, I believe?

blychs:
> @blychs did you have any thoughts about these pooch comments [from](https://github.com/NOAA-CSL/MELODIES-MONET/pull/326#discussion_r1953478536) @DWesl?
> 
> > It doesn't look like pooch has a way to set the file name after the download has started, but you might be able to guess from the URL. Unfortunately, there's enough variation in how the URL encodes information about the resource it points to that a universal method of composing a simple filename obviously related to the URL is probably not possible.
> > Pooch's default involves a hash of the URL, would using just the hash without the guess at the filename work? It's not as obviously related to the URL if you want to find the downloaded file in pooch's cache, but the code will be able to reconstruct it easily. It might be worth adding some part of the domain/hostname in the hash to get a little association between URL and cached file name and to reduce the chance of filename collisions.

I just answered under that topic. I think the idea is quite nice, but I don't find an easy way to make it work, because I believe that we need to have the extension of the file. Might be wrong though, or there might be a way to infer it.

zmoon:
@blychs If you can merge upstream develop into your branch the file header check should pass, if you add an additional `#` line (see [the recent change](https://github.com/NOAA-CSL/MELODIES-MONET/pull/332/files#diff-9373eef859a5e8518167dd56d0895961ec1e29d478477946c7fac611ce453b4e)).

blychs:
Done

DWesl:
> I just answered [under that topic](https://github.com/NOAA-CSL/MELODIES-MONET/pull/326#discussion_r1955020845). I think the idea is quite nice, but I don't find an easy way to make it work, because I believe that we need to have the extension of the file. Might be wrong though, or there might be a way to infer it.

That's a pooch error then?  In that case, parsing through `fmt=shp` vs. `format=shapefile` vs. `name=file.shp` in the parameters vs an actual filename before the parameters would be a never-ending pain.

You could create and update a mapping of URL to filename if you wanted to get back some of the caching pooch does, if you take steps to avoid filename collisions.  (If you're not able to persist the mapping between script invocations, you could set up an `atexit` handler to delete all the files when Python exits).

If you are already using this as a fallback filename:
```python
if fname is None:
    fname = url.rsplit("/", 1)[-1]
```
changing to have `fname = os.path.basename(urllib.parse.urlparse(url).path)` might produce a normal filename in more cases, but it might also get rid of information you want to keep .  I don't know how much of a concern weird non-descriptive filenames would be, but `os.path.basename(url)` might be a bit more clear about what you're doing if you want to keep the current approach.  (Pooch's default would then be something like `"-".join([hashlib.sha1(url.encode('utf-8')).hexdigest(), os.path.basename(urllib.parse.urlparse(url).path)])` or `"-".join([hashlib.sha1(url.encode('utf-8')).hexdigest(), os.path.basename(url)])`, which are less likely to overwrite other files from the same source, but if you always re-download the file that's less relevant)

You could also do this assignment unconditionally before the Content-Disposition check and overwrite it if the response indicates a preferred filename so `fname` always has a string value.

DWesl:
> Sadly, that is not the problem. The problem is just that some shapefiles, like the one in the example, don't have the name of the file in the URL at all. It is still available in content-disposition. Of course, we could tell the user to define the name in domain_info, but I thin that having it be automatic is probably better.

If you are not saving the downloaded file path somewhere for use later, you could just merge `mask_path` and `mask_url` to pass whichever is passed to [`geopandas.read_file`](https://geopandas.org/en/stable/docs/reference/api/geopandas.read_file.html).

If you knew the format would always be a simple shapefile, you set the pooch name to `"regions.shp"` or `"-".join([hashlib.sha1(url).hexdigest(), "regions.shp"])`, but there is an excellent chance that shapefiles will be `.shz` or whatever their bundled files are called instead.  

If pooch just needs a name and ignores extensions, you could tell it the name is `hashlib.sha1(url).hexdigest()` and hope nobody needs to work out where entries in the pooch cache came from

blychs:
I am not sure I am understanding the suggestion correctly.
Currently, I am indeed saving the file so that it can be used later. 
Pooch does indeed download it, but the name is a bunch of gibberish because it is determined with the URL.
The solution I wrote, and @zmoon made better, bases the file name in the content-disposition. Having proper file names helps to save them for future use.

DWesl:
So the sensible local file names (with extension) are an important part of the function (well, as sensible as they get: some sites will name all such files `download.zip`).

You are saving the file, but re-running the script would download it again, as would using the same URL later in the same script.  One would have to examine the run directory to determine the name of the downloaded file (and hope that name is descriptive enough if there are multiple such URLs in the script that you can do so) then change the function call to use the discovered path rather than the URL, unless I am missing something?

I think, with your current API, applying `functools.lru_cache` to `_download_file` would take care of multiple uses of the same URL in a single script.

For the second, I think you might need an explicit cache to save to disk on script exit and read back if present on script start, something like
```python
_CACHE_FILE = "regionmask-downloaded-file-names.pickle"
try:
    with open(_CACHE_FILE, "rb") as in_file:
        _URL_TO_NAME = pickle.load(in_file)
        # type: dict[str, str]
except FileNotFoundError:
    _URL_TO_NAME = {}

def _download_with_name(url: str) -> str:
    """Download file from URL, using a local copy if available."""
    try:
        return _URL_TO_NAME[url]
    except KeyError:
        pass
    fname = os.path.basename(urllib.parse.urlparse(url).path)
    # download file
    ...
    if content_disposition:
        # extract suggested file name
        fname = ...
    if fname in _URL_TO_NAME.values():
        fname = "-".join([hashlib.sha1(url).hexdigest(), fname])
    _URL_TO_NAME[url] = fname
    # save file to disk
    ...
    return fname

def _save_url_cache():
    """Save the url-to-fname cache to disk."""
    with open(_CACHE_FILE, "wb") as out_file:
        pickle.dump(_URL_TO_NAME, out_file)

atexit.register(_save_url_cache)
```
since something like this might come up every so often:
```pycon
>>> os.path.basename(urllib.parse.urlparse("https://js.arcgis.com/path/to/conus_states.shp?datum=NAD83").path)
'conus_states.shp'
```

This cache isn't quite thread-safe, but I don't think that's relevant here.

blychs:
I agree that thread safety doesn't really matter here.
If I understand your script correctly, the idea is to save a cache of the filenames and URL's, to avoid duplicate downloads. To me, it seems like an acceptable tradeof. If @zmoon and/or @rschwant agree, we can move forward with that.
Cheers,
Pablo

zmoon:
I think, as a user, if I were to specify a shapefile with a URL, I would *expect* it to be downloaded at least once when MM runs. The docs can make it clear.

@DWesl mentions `functools.lru_cache`. What about simplifying things with something like this?
```python
from functools import lru_cache

import geopandas as gpd
import regionmask

@lru_cache(None)
def get_regions(url_or_path, **kwargs):
    return regionmask.from_geopandas(gpd.read_file(url_or_path), **kwargs)
```
<details><summary>Example with timing</summary>

```
In [2]: kwargs = dict(names="CONTINENT", abbrevs="_from_name", name="continent")
   ...: url = "https://pubs.usgs.gov/of/2006/1187/basemaps/continents/continents.zip"

In [3]: %time get_regions(url, **kwargs)
CPU times: user 425 ms, sys: 90 ms, total: 515 ms
Wall time: 13.6 s
Out[3]:
<regionmask.Regions>
Name:     continent
overlap:  False

Regions:
0    Asi          Asia
1 NorAme North America
2    Eur        Europe
3    Afr        Africa
4 SouAme South America
5    Oce       Oceania
6    Aus     Australia
7    Ant    Antarctica

[8 regions]

In [4]: %time get_regions(url, **kwargs)
CPU times: user 4 µs, sys: 0 ns, total: 4 µs
Wall time: 6.68 µs
Out[4]:
<regionmask.Regions>
Name:     continent
overlap:  False

Regions:
0    Asi          Asia
1 NorAme North America
2    Eur        Europe
3    Afr        Africa
4 SouAme South America
5    Oce       Oceania
6    Aus     Australia
7    Ant    Antarctica

[8 regions]
```
</details>

I also don't think this discussion has to hold back this PR. We can always continue it elsewhere (GitHub Issue or Discussion), and develop this functionality more in the future (and perhaps add some pytest cases, etc.).

blychs:
That seems good to me. Agree to add that, test this, merge this and then continue the discussion elsewhere to make it better.

blychs:
> I think, as a user, if I were to specify a shapefile with a URL, I would _expect_ it to be downloaded at least once when MM runs. The docs can make it clear.
> 
> @DWesl mentions `functools.lru_cache`. What about simplifying things with something like this?
> 
> ```python
> from functools import lru_cache
> 
> import geopandas as gpd
> import regionmask
> 
> @lru_cache(None)
> def get_regions(url_or_path, **kwargs):
>     return regionmask.from_geopandas(gpd.read_file(url_or_path), **kwargs)
> ```
> 
> Example with timing
> 
> I also don't think this discussion has to hold back this PR. We can always continue it elsewhere (GitHub Issue or Discussion), and develop this functionality more in the future (and perhaps add some pytest cases, etc.).

I was just testing it, and it doesn't seem to work in the case of URLs having random names (so, it still has the problem of content-disposition). If @zmoon  and @DWesl agree, I would move forward with merging the original, _too many downloads_ version and keep testing other, better variants. I am happy to add it as a todo/issue.

Cheers
Pablo

zmoon:
@blychs what's an example of one that doesn't work with that? I tried your one from the YAML doc.

blychs:
@zmoon Maybe I did something wrong when copying your code snippet? I tried the one I added to the docs: https://hub.arcgis.com/api/v3/datasets/66c2642209684b90af84afcc559a5a02_5/downloads/data?format=shp&spatialRefId=4269&where=1%3D1
This has the Colorado Counties. I added an example to the docs with Denver county

DWesl:
What sort of errors are you seeing?  With the posted code and no kwargs, I can get regionmask to get me a list of regions named "Region0" to "Region63".  With a bit of exploring, I can convince it to use the county names, but guessing those keyword arguments required exploring the dataframe a bit.  

Perhaps this amended function would help witht that exploring?
```python
@lru_cache(None)
def get_regions(url_or_path, verbose=True, **kwargs):
    df = gpd.read_file(url_or_path)
    if verbose:
        print(df.columns)
    return regionmask.from_geopandas(df, **kwargs)
```

Which versions of regionmask and geopandas are you using?

blychs:
> What sort of errors are you seeing? With the posted code and no kwargs, I can get regionmask to get me a list of regions named "Region0" to "Region63". With a bit of exploring, I can convince it to use the county names, but guessing those keyword arguments required exploring the dataframe a bit.
> 
> Perhaps this amended function would help witht that exploring?
> 
> ```python
> @lru_cache(None)
> def get_regions(url_or_path, verbose=True, **kwargs):
>     df = gpd.read_file(url_or_path)
>     if verbose:
>         print(df.columns)
>     return regionmask.from_geopandas(df, **kwargs)
> ```
> 
> Which versions of regionmask and geopandas are you using?

That is likely the explanation. Mine breaks because it cannot detect the counties (I get a KeyError). I have regionmask 0.12.1 and geopandas 1.0.1. It might just require more tinkering to do that right.

DWesl:
`get_regions(url, abbrevs="LABEL", names="FULL", numbers="NUM_FIPS", name="us-colorado-counties")` was what ended up working, and it looks like you have that information in `region_selection.rst`: did you pass those arguments through to `get_regions`?

Coming at this from the other direction, what kind of errors do you see if you drop the ZIP-file handling on lines 169-170?  I don't see a difference in what happens if I read `file.zip` or `zip://file.zip` (I do see problems if I rename it `data` and try to read that, with or without the `zip://` leader).

blychs:
It works perfectly, I had messed up my test.

blychs:
I changed it to @zmoon 's. I still there is still value in having the user disntinguish between url's and file paths, but if you think it's unnecessary, with the new function that could also be eliminated. I am also happy to add @DWesl checks if we think it can help catch some errors.
Cheers
Pablo

DWesl:
Which checks would those be?  The ones I remember are to avoid saving multiple files to the same path, which you could do with `os.path.exists` and a warning, if the code still writes the download to a file, which I don't think @zmoon's `get_regions` does.
Printing the column names of the shapefile before passing to regionmask would be helpful in getting useful labels, or perhaps the first two rows (`df.head(2)`)

And because I was tickled by the problem of guessing the file name corresponding to a download, here's functions to guess based on URL and HTTP headers, using only stdlib functions: [guess_download_file_name.py](https://github.com/user-attachments/files/18919093/guess_download_file_name.py.txt)

If you decide to go for a persistent cache later, it might be worth adding `r.headers.ETag` with the saved file name, to help with cache invalidation (many servers, when told when we last downloaded a file and what version/ETag we downloaded, will tell us what we have is fine rather than sending the whole thing again if there aren't updates, and we can get download time from `os.stat(path).st_mtime`).  I also noticed the stdlib [`shelve`](https://docs.python.org/3/library/shelve.html) module might be handy, but that's also for later.

melodies_monet/util/region_select.py
```diff
@@ -201,7 +201,7 @@ def control_custom_mask(data, domain_type, domain_info=None, **kwargs):
     else:
         raise ValueError(
             "Could not identify the type of domain. Should be 'polygon',"
-            + " 'defined-region' or 'file'"
+            + f" 'defined-region' or 'file'. You asked for {domain_type}"
```
zmoon, line 204:
```suggestion
            + f" 'defined-region' or 'file'. You asked for {domain_type!r}"
```

blychs:
I think that now it's ready to merge @rschwant

PR #327, opened 2025-01-14
Feature/calculate columns

blychs:
Adds function to calculat partial/total columns and local geographic time in tools.py, since this is used by TEMPO, TROPOMI, and will be used by Pandora and others, and right now it was coded over and over again.


blychs:
Closed because I created the pool request to main instead of develop

PR #328, opened 2025-01-14
Feature/calculate columns

blychs:
Adds function to calculat partial/total columns and local geographic time in tools.py, since this is used by TEMPO, TROPOMI, and will be used by Pandora and others, and right now it was coded over and over again.
Also, move import xarray to the top. It's called by many functions and that is the recommended practice according to PEP8. I don't see any good reason to not have it at the top.

### DIFF
--- a/melodies_monet/util/tools.py
+++ b/melodies_monet/util/tools.py
@@ -555,7 +555,7 @@ def calc_totalcolumn(modobj, var="NO2"):
     """
     data = calc_partialcolumn(modobj, var)
     try:
-        data = data.where(modobj['pres_pa_mid'] > modobj['surfpres_pa'])
+        data = data.where(modobj['pres_pa_mid'] >= modobj['surfpres_pa'])
     except KeyError:
         pass
     total_col = data.sum(dim='z', keep_attrs=True)
--- a/melodies_monet/util/tools.py
+++ b/melodies_monet/util/tools.py
@@ -344,8 +344,8 @@ def mobile_and_ground_pair(ds_model,df_obs, var_name_list):
     df_model.drop(labels=['x','y','time_obs'], axis=1, inplace=True)
 
     final_df_model = merge_asof(df_obs, df_model, 
-                            by=['latitude', 'longitude'], 
-                            on='time', direction='nearest')
+                            by=['latitude', 'longitude'],
+                            on='time', direction='nearest', suffixes=('', '_new'))
 
     return final_df_model
 
--- a/melodies_monet/util/tools.py
+++ b/melodies_monet/util/tools.py
@@ -6,6 +6,7 @@ from __future__ import division
 from builtins import range
 
 import numpy as np
+import xarray as xr
 
 __author__ = 'barry'
 
@@ -277,7 +278,6 @@ def get_epa_region_df(df):
 
 def resample_stratify(da, levels, vertical, axis=1,interpolation='linear',extrapolation='nan'):
     import stratify
-    import xarray as xr
 
     result = stratify.interpolate(levels, vertical.chunk().data, da.chunk().data, axis=axis,
                                  interpolation = interpolation,extrapolation = extrapolation)
@@ -294,7 +294,6 @@ def resample_stratify(da, levels, vertical, axis=1,interpolation='linear',extrap
     return out
 
 def vert_interp(ds_model,df_obs,var_name_list):
-    import xarray as xr
     from pandas import merge_asof, Series
 
     var_out_list = []
@@ -313,13 +312,7 @@ def vert_interp(ds_model,df_obs,var_name_list):
         var_out_list.append(out)
 
     df_model = xr.merge(var_out_list).to_dataframe().reset_index()
-    for time in df_model.time.unique():
-        if df_model[df_model.time == time].pressure_obs.unique() > df_model[df_model.time == time].pressure_model.max():
-            df_model.fillna({'pressure_model':df_model[df_model.time == time].pressure_obs},inplace=True)
-        elif df_model[df_model.time == time].pressure_obs.unique() < df_model[df_model.time == time].pressure_model.min():
-            df_model.fillna({'pressure_model':df_model[df_model.time == time].pressure_obs},inplace=True)
-            print('Warning: You are pairing obs data above the model top. This is not recommended.')
-            print(time)
+    df_model.fillna({'pressure_model':df_model.pressure_obs},inplace=True)
     df_model.drop(labels=['x','y','z','pressure_obs','time_obs'], axis=1, inplace=True)
     df_model.rename(columns={'pressure_model':'pressure_obs'}, inplace=True)
 
@@ -330,7 +323,6 @@ def vert_interp(ds_model,df_obs,var_name_list):
     return final_df_model
 
 def mobile_and_ground_pair(ds_model,df_obs, var_name_list):
-    import xarray as xr
     from pandas import merge_asof, Series
     
     var_out_list = []
@@ -353,7 +345,7 @@ def mobile_and_ground_pair(ds_model,df_obs, var_name_list):
 
     final_df_model = merge_asof(df_obs, df_model, 
                             by=['latitude', 'longitude'], 
-                            on='time', direction='nearest', suffixes=('', '_new'))
+                            on='time', direction='nearest')
 
     return final_df_model
 
@@ -376,7 +368,6 @@ def find_obs_time_bounds(files=[],time_var=None):
     """
     import os 
     import monetio as mio
-    import xarray as xr
     
     if isinstance(files,str):
         files = [files]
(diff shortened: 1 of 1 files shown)


blychs:
Hey all, I would like to use this PR for the Pandora total column comparisons. This means that the faster it can be reviewed (and hopefully any bugs fixed!), the better. If it is, we could also update the other satellite tools to use this, and reduce code duplication.

melodies_monet/util/tools.py
```diff
@@ -353,7 +345,7 @@ def mobile_and_ground_pair(ds_model,df_obs, var_name_list):
 
     final_df_model = merge_asof(df_obs, df_model, 
                             by=['latitude', 'longitude'], 
-                            on='time', direction='nearest', suffixes=('', '_new'))
```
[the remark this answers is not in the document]
blychs: This is a mistake, probably wrongly reverted in a merge. I'll undo this

melodies_monet/util/tools.py
```diff
@@ -508,3 +499,76 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+    var : str
+        variable to calculate the partial column from
+
+    Returns
+    -------
+    xr.DataArray
+        DataArray containing the partial column of the species.
+    """
+    R = 8.314  # m3 * Pa / K / mol
+    NA = 6.022e23
```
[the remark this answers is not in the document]
blychs: RIght now we are not importing scipy, and the difference is unlikely to matter taking into account normal model errors. I am a little bit hesitant about adding new imports, but if you don't think it's a problem, I'm happy to do so. I could also just create the constants at the beginning of the file with larger precision, if it's worth it but we still prefer not to import SciPy.

melodies_monet/util/tools.py
```diff
@@ -508,3 +499,76 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+    m2_to_cm2 = 1e4
+    fac_units = ppbv2molmol * NA / m2_to_cm2
+    partial_col = (
+        modobj[var]
+        * modobj["pres_pa_mid"]
+        * modobj["dz_m"]
+        * fac_units
+        / (R * modobj["temperature_k"])
+    )
+    partial_col.attrs = {"units": "molecules/cm2", "description": f"{var} partial column"}
```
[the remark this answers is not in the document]
blychs: I'm always confused about how the long name and the description should be different. I just put it as "long_name": f"partial column of {var}", but am not sure whether that solves the issue?

melodies_monet/util/tools.py
```diff
@@ -508,3 +499,76 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+    var : str
+        variable to calculate the total column from
+
+    Returns
+    -------
+    xr.DataArray
+        DataArray containing the total column of the species.
+    """
+    data = calc_partialcolumn(modobj, var)
+    return data.sum(dim='z', keep_attrs=True)
```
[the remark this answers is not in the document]
blychs: Done

melodies_monet/util/tools.py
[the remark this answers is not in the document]
zmoon: Well one aspect is that `long_name` is what xarray prefers to use for plot labels so you usually you want it to be reasonably concise. `description` could be longer with more details, or just left off.

melodies_monet/util/tools.py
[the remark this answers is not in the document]
blychs: I defined the constants at the beginning of the module and used them along it, which also improves readability (IMHO).
While doing this, I noticed that `convert_std_amb_bc` and `convert_std_amb_ams` seem to be actually the exact same functions?

melodies_monet/util/tools.py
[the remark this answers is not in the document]
blychs: Replaced `description` with `long_name` based on this.

melodies_monet/util/tools.py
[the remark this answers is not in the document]
zmoon: SciPy is included in monetio conda-forge deps, so an MM install should always have it. So I wouldn't feel bad (though we would want to update MM deps anyway). I think the way you did it is fine though, and having more precision for these constants won't hurt.

melodies_monet/util/tools.py
```diff
@@ -313,13 +315,7 @@ def vert_interp(ds_model,df_obs,var_name_list):
         var_out_list.append(out)
 
     df_model = xr.merge(var_out_list).to_dataframe().reset_index()
-    for time in df_model.time.unique():
-        if df_model[df_model.time == time].pressure_obs.unique() > df_model[df_model.time == time].pressure_model.max():
-            df_model.fillna({'pressure_model':df_model[df_model.time == time].pressure_obs},inplace=True)
-        elif df_model[df_model.time == time].pressure_obs.unique() < df_model[df_model.time == time].pressure_model.min():
-            df_model.fillna({'pressure_model':df_model[df_model.time == time].pressure_obs},inplace=True)
-            print('Warning: You are pairing obs data above the model top. This is not recommended.')
-            print(time)
```
blychs, line 322: I hadn't noticed this, which seems like another mistake. I do not recall changing this myself, and it might be me not pulling correctly. Should I undo this?

zmoon: If you didn't commit that deletion yourself somewhere, you should be able to `git merge` the upstream develop branch into your branch to fix this.

zmoon: Seems like you did though in https://github.com/blychs/MELODIES-MONET/commit/a5c28f927630dff036e8878fea7636cc9f5e83b2 so maybe you'll have to fix it manually

melodies_monet/util/tools.py
```diff
@@ -498,13 +492,85 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+    var : str
+        variable to calculate the total column from
+
+    Returns
+    -------
+    xr.DataArray
+        DataArray containing the total column of the species.
+    """
+    data = calc_partialcolumn(modobj, var)
+    total_col = data.sum(dim='z', keep_attrs=True)
```
DWesl, line 551: Should this start from surface pressure (e.g. `data.where("air_pressure > surface_air_pressure").sum(dim="z", keep_attrs=True)`), or is that handled elsewhere (the previous function implies the vertical coordinate is in meters, but I don't recall if that is height above ground or altitude measured from sea level)?

blychs: It is not really handled, all of the previous functions were assuming that the highest pressure is the surface_pressure. If that's not the case, this will be wrong, so the check makes sense. I am not sure whether all of the models include the surface_pressure though?

melodies_monet/util/tools.py
```diff
@@ -498,13 +498,90 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+        variable to calculate the total column from
+
+    Returns
+    -------
+    xr.DataArray
+        DataArray containing the total column of the species.
+    """
+    data = calc_partialcolumn(modobj, var)
+    try:
+        data = data.where(modobj['pres_pa_mid'] >= modobj['surfpres_pa'])
```
DWesl, line 558:
```suggestion
        data = data.where(modobj['pres_pa_mid'] <= modobj['surfpres_pa'])
```
I think it's currently only using underground values, rather than only overground values (tropopause pressure ~100 hPa, surface pressure ~ 1000 hPa, and yes, I always have to think about whether a given vertical coordinate increases with height)

blychs: Of course 🤦

blychs:
Is there anything missing to merge this one?

melodies_monet/util/tools.py
```diff
@@ -498,13 +498,90 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
     
     # amb to std = Losch / Airnum
     convert_std_to_amb_bc = Airnum/std_bc
     
     for var in convert_vars:
         ds[var] = ds[var]*convert_std_to_amb_bc
 
+
+def calc_partialcolumn(modobj, var="NO2"):
+    """Calculates the partial column of a species from its concentration.
```
DWesl, line 513: Perhaps clarify that this is only the part of the column within each grid cell?  I've been working with literature where the partial column refers to the integral from the top of the atmosphere down to that point.  Either a new paragraph before the Parameters section, or an additional line in the Returns section would work

melodies_monet/util/tools.py
```diff
@@ -498,13 +498,90 @@ def convert_std_to_amb_bc(ds,convert_vars=[],temp_var=None,pres_var=None):
...
+    Returns
+    -------
+    xr.DataArray
+        DataArray containing the local time based on longitude.
+    """
+    # Make sure that lon is in the range [-180, 180]
+    # This should be guaranteed by the reader, and it isn't needed,
+    # but it is very cheap to redo and should make us be safer.
+
+    hrs2ms = 3600000
```
DWesl, line 583:
```suggestion
    hrs2ms = 3600_000
```
Might help with clarity and with checking the number of zeros at a glance ([new in Python 3.6](https://docs.python.org/3/whatsnew/3.6.html#pep-515-underscores-in-numeric-literals), if that's relevant to support considerations)

melodies_monet/util/tools.py
DWesl: Thank you, looks good.  I remember the difference for ozone was a similar order of magnitude to the ozone hole, though I don't have numbers on hand.

zmoon (approved):
LGTM

PR #330, opened 2025-01-28
Feature: rename observation coordinates for xarray datasets

mbruckner-work:
While testing if the existing aircraft tools would work for IAGOS observations I discovered that the existing rename_vars function in the obs class does not allow renaming of xarray coordinate variables through the yaml variables option. The IAGOS files are netcdf. When these are opened with xr.open_dataset lat, lon, time, and an altitude variable are assigned as coordinates. The coordinate names aren't the monetio standard names.

I was able to determine that for xarray datasets the 'data_vars' attribute does not include the names variables labeled as coordinates. This pull request adds the coordinate variable names (eg. lat, lon, time) to the data variable name list when the obs dataset is an xarray.Dataset.


colin-harkins (approved):
This looks good to me, thanks for addressing this!

PR #340, opened 2025-03-03
CLI: Get `get-openaq` working again (web API v3)

zmoon:
This uses https://github.com/noaa-oar-arl/monetio/pull/235

cc: @jordanschnell 

An example for testing. EC (Ecuador) has 10 sites, 8 of which are active for PM2.5 during this period.
```
melodies-monet get-openaq -s 2021-01-01 -e 2021-09-30 -p pm25 -c EC --reference-grade --low-cost --verbose --debug
```

One of the new things in v3 is sensor IDs in addition to location IDs (sites). Technically there could be multiple sensors for the same parameter at one site-time. We could handle that here, or in the reader's long->wide format conversion.

### DIFF
--- a/melodies_monet/_cli.py
+++ b/melodies_monet/_cli.py
@@ -1238,14 +1238,20 @@ def get_openaq(
             "Parameters. "
             "Use '-p' more than once to get multiple parameters. "
             "Other examples: 'no', 'no2', 'nox', 'so2', 'co', 'bc'. "
-            "Only applicable to the web API methods ('api-v2')."
+            "Only applicable to the web API methods ('api-v*')."
         )
     ),
     reference_grade: bool = typer.Option(True, help="Include reference-grade sensors."),
     low_cost: bool = typer.Option(False, help="Include low-cost sensors."),
-    method: str = typer.Option("api-v2", "-m", "--method", help=(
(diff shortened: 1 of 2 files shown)


melodies_monet/_cli.py
```diff
@@ -1409,7 +1447,25 @@ def get_openaq(
...
+            site_vns = [
+                "siteid",  # real OpenAQ location ID
+                "latitude",
+                "longitude",
+                "utcoffset",
+                #
+                "country",
+                #
+                "is_mobile",
+                "is_monitor",
```
zmoon, line 1467: In the previous version, this was encompassed in a single string variable, `sensor_type`. We could keep that behavior here. That's probably the only thing that might change in this PR. Other potential changes, like for the rate limiting, would be just on the monetio side.

rschwant: I think it's okay to just leave the labels to be consistent with OpenAQ version 3 here and we do not have to keep the old behavior.

rschwant (approved):
Looks great! As discussed let's get this in for Version 1 and potentially there are updates on the MONETIO side which can get in after more verification from Jordan.

jordanschnell (approved):

PR #348, opened 2025-03-18
Update mopitt monthly/daily check

mbruckner-work:
While testing with multiple days of UFS-AQM data it became apparent that the handling of switching for monthly vs daily MOPITT observation data wasn't quite right. Obs.time and model.time size comparisons were old way. 

Not completely sure how it didn't pop up with RAQMS-based testing.

### DIFF
--- a/melodies_monet/util/satellite_utilities.py
+++ b/melodies_monet/util/satellite_utilities.py
@@ -110,14 +110,8 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname,global_model=True):
         raise
     
     # Aggregate time-step, if needed
-    ## Check if same number of timesteps:
-    if obs_data.time.size == model_data.time.size:
-        model_obstime = model_data
-        filtstr = '%Y-%m'
-    elif obs_data.time.size > model_data.time.size:
-        print('Observation data appears to be a finer time resolution than model data')
-        raise
-    elif obs_data.attrs['monthly']:
+    ## Check if obs are monthly or daily
+    if obs_data.attrs['monthly']:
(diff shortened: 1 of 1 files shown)


melodies_monet/util/satellite_utilities.py
```diff
@@ -110,14 +110,8 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname,global_model=True):
         raise
     
     # Aggregate time-step, if needed
-    ## Check if same number of timesteps:
-    if obs_data.time.size == model_data.time.size:
-        model_obstime = model_data
-        filtstr = '%Y-%m'
-    elif obs_data.time.size > model_data.time.size:
-        print('Observation data appears to be a finer time resolution than model data')
```
rrbuchholz, line 118: I can't remember - could this be raised if we have just accidentally loaded a longer dataset from the model? For example defined different time bounds in the yaml? Or does the processing apply a date start and date end cut first?

melodies_monet/util/satellite_utilities.py
```diff
@@ -110,14 +110,8 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname,global_model=True):
...
-        raise
-    elif obs_data.attrs['monthly']:
+    ## Check if obs are monthly or daily
+    if obs_data.attrs['monthly']:
```
rrbuchholz, line 114: So it looks like we have taken out the dependency of the model dataset being larger than the observational dataset -- I think this is a good move, because then we just match dates specifically rather than assuming the same number of points means the same dates.

rrbuchholz (approved):
I think the changes look good to me.

PR #350, opened 2025-03-20
generalization of TROPOMI for additional models

mbruckner-work:
Modifies tropomi NO2 pairing code developed by Meng to work for models like UFS-AQM or RAQMS. 

- Incorporates sampling of model to overpass time as done for MOPITT
- Potential use of model tropopause pressure in calculating model tropospheric column. Requires to be named 'pres_pa_trop'. 
- Removes hardwire of model no2 variable name.
- Rewrote portions to use xarray.


melodies_monet/driver.py
```diff
@@ -590,6 +590,7 @@ def open_model_files(self, time_interval=None, control_dict=None):
             self.mod_kwargs.update({"fname_met_2D": control_dict['model'][self.label].get('files_met_surf', None)})
             self.obj = mio.models._camx_mm.open_mfdataset(self.files, **self.mod_kwargs)
         elif 'raqms' in self.model.lower():
+            self.mod_kwargs.updata({'var_list': list_input_var})
```
blychs, line 593:
```suggestion
            self.mod_kwargs.update({'var_list': list_input_var})
```
Do you mean update?

melodies_monet/driver.py
```diff
@@ -1364,10 +1360,11 @@ def pair_data(self, time_interval=None):
                             
                             paired_data = sutil.omps_nm_pairing_apriori(model_obj,obs.obj,keys)
                         else:
-                            model_obj = mod.obj[keys+['dp_pa']]
+                            keys.extend('dp_pa')
+                            model_obj = mod.obj[keys]
                             paired_data = sutil.omps_nm_pairing(model_obj,obs.obj,keys)
 
-                        paired_data = paired_data.where((paired_data.o3vmr > 0))
+                        paired_data = paired_data.where((paired_data.ozone_column > 0))
```
blychs, line 1367: I had been told that, as weird as it sounds, we should not discard negative values. Might be wrong though.
Could it be left to the filter options for the user? Or is there something in the reader that makes something like that difficult?

blychs (commented):
Great job!

blychs:
Another issue that I'm getting.
region_select assumes that latitude and longitude are called `latitude`"and `longitude`. However, in the pair TROPOMI data, while the data variables are `latitude` and `longitude`, the coords are called `lat` and `lon`. This inconsistency makes `region_select` break.
Two possible fixes:

- Option 1: Guarantee that they are called `latitude` and `longitude`, either at the `MONETIO` level or in `MELODIES-MONET`.
- Option 2: add a check to `region_select`.

blychs (changes requested):
TROPOMI paired data uses `latitude` and `longitude` as variables, but `lat` and `lon` as coords. This causes `region_select` to crash when `regionmask` is required.
Two possible fixes:

- Option 1 (I lean towards this one): rename TROPOMI paired coords to `latitude` and `longitude`, consistent with the other sat tools.
- Option 2: Add a check to `region_select`

mbruckner-work:
> TROPOMI paired data uses `latitude` and `longitude` as variables, but `lat` and `lon` as coords. This causes `region_select` to crash when `regionmask` is required. Two possible fixes:
> 
> * Option 1 (I lean towards this one): rename TROPOMI paired coords to `latitude` and `longitude`, consistent with the other sat tools.
> * Option 2: Add a check to `region_select`

I'm a little confused by this as the new version has the coords `latitude` and `longitude`. I thought that I changed the structure of the paired data to drop having latitude/longitude in both variables and coords with different names...

blychs:
> > TROPOMI paired data uses `latitude` and `longitude` as variables, but `lat` and `lon` as coords. This causes `region_select` to crash when `regionmask` is required. Two possible fixes:
> > 
> > * Option 1 (I lean towards this one): rename TROPOMI paired coords to `latitude` and `longitude`, consistent with the other sat tools.
> > * Option 2: Add a check to `region_select`
> 
> I'm a little confused by this as the new version has the coords `latitude` and `longitude`. I thought that I changed the structure of the paired data to drop having latitude/longitude in both variables and coords with different names...

Let me check, I might have updated my version of your code wrongly.

blychs:
While running this with your branc, it complains about the variable NO2_col missing (NO2 is the name in CAMx, the model I'm testing with). Is it possible that there is an uncommitted change?

melodies_monet/driver.py
mbruckner-work: This part is for OMPS, where I needed to handle a bit better where there wasn't actually any obs data. From the obs files, there is a valid range of [ 50. 700.] for the total column O3. 
How about a discard of where the obs is np.nan?

```suggestion
                        paired_data = paired_data.where(~np.isnan(paired_data.ozone_column))
```

blychs: Looks good to me. If you think that keeping the `>0` option is better, I don't have a strong opinion anyway

blychs:
It is working now, adding the `mod_to_overpass` keywords as you suggested by email. I don't know why, it does not do the states and counties in the maps, but that's probably a satplots issue and not related to the PR. It's weird, though, because that should be handled at the `MONET` level.
Of course, we also have the lingering satplots issue with timeseries (they are completely incorrect), but I don't think that's necessarily related to this PR either.

docs/develop/development_team.rst
```diff
@@ -26,7 +26,7 @@ David Fillmore     dwfncar         Satellite obs, Plots, Stats, Connection to ME
 Colin Harkins      colin-harkins   Docs, Aircraft obs, Plots, Stats
 Louisa Emmons      lkemmons        Docs, Aircraft obs, Plots, Stats
 Rebecca Schwantes  rschwant        Docs, Surface and aircraft obs, Plots, Stats
-Margaret Bruckner  mbruckner-work  Docs, Satellite obs, Plots, Stats
+Maggie Bruckner  mbruckner-work  Docs, Satellite obs, Plots, Stats
```
blychs, line 29:
```suggestion
Maggie Bruckner    mbruckner-work  Docs, Satellite obs, Plots, Stats
```
I believe the automatic test is failing just because of the alignment to build the docs

mbruckner-work:
@blychs I think I accidentally introduced that issue when I was merging in the develop branch. Will update to your suggestion.

blychs:
Did you decide what to do with mod_to_overpass? If you are convinced that the best is leaving it as is, I'm OK with approving your PR, but maybe adding the error message would be a good idea.

mbruckner-work:
> Did you decide what to do with `mod_to_overpass`? If you are convinced that the best is leaving it as is, I'm OK with it, but maybe adding the error message would be a good idea.

Decided to add an else with a printed warning and partial column calculation.

blychs (approved):

melodies_monet/driver.py
```diff
@@ -600,14 +601,12 @@ def open_model_files(self, time_interval=None, control_dict=None):
...
                 self.obj = mio.models.raqms.open_dataset(file_list)
+            self.obj = self.obj.rename({'ptrop':'pres_pa_trop'})
 
         else:
             print('**** Reading Unspecified model output. Take Caution...')
-            if len(self.files) > 1:
+            if len(self.files) >= 1:
                 self.obj = xr.open_mfdataset(self.files,**self.mod_kwargs)
-            else:
-                self.obj = xr.open_dataset(self.files[0],**self.mod_kwargs)
```
zmoon, line 609: I would put this block back the way it was. Usually no need to get mfdataset for one file.

melodies_monet/driver.py
```diff
@@ -600,14 +601,12 @@ def open_model_files(self, time_interval=None, control_dict=None):
                 self.obj = mio.models.raqms.open_mfdataset(file_list,**self.mod_kwargs)
             else:
                 self.obj = mio.models.raqms.open_dataset(file_list)
+            self.obj = self.obj.rename({'ptrop':'pres_pa_trop'})
```
zmoon, line 604: Should we do this in the reader instead?

For compat, you could change it to this:
```suggestion
            if 'ptrop' in self.obj and 'pres_pa_trop' not in self.obj:
                self.obj = self.obj.rename({'ptrop':'pres_pa_trop'})
```

melodies_monet/driver.py
```diff
@@ -1364,10 +1360,11 @@ def pair_data(self, time_interval=None):
                             
                             paired_data = sutil.omps_nm_pairing_apriori(model_obj,obs.obj,keys)
                         else:
-                            model_obj = mod.obj[keys+['dp_pa']]
+                            keys.extend('dp_pa')
+                            model_obj = mod.obj[keys]
                             paired_data = sutil.omps_nm_pairing(model_obj,obs.obj,keys)
 
-                        paired_data = paired_data.where((paired_data.o3vmr > 0))
+                        paired_data = paired_data.where(~np.isnan(paired_data.ozone_column))
```
zmoon, line 1367:
```suggestion
                        paired_data = paired_data.where(paired_data.ozone_column.notnull())
```

melodies_monet/driver.py
```diff
@@ -1364,10 +1360,11 @@ def pair_data(self, time_interval=None):
                             
                             paired_data = sutil.omps_nm_pairing_apriori(model_obj,obs.obj,keys)
                         else:
-                            model_obj = mod.obj[keys+['dp_pa']]
+                            keys.extend('dp_pa')
+                            model_obj = mod.obj[keys]
                             paired_data = sutil.omps_nm_pairing(model_obj,obs.obj,keys)
```
zmoon, line 1365: 'dp_pa' is added within this function too, maybe remove that. Or revert your change here.

zmoon: And it looks like the function is expecting a single string to be passed, based on the arg name `ozone_ppbv_varname` and `pair_variables = ['dp_pa', ozone_ppbv_varname]`. Some fixes are needed (decide whether to pass single string var or list of strings, and adjust the function accordingly).

melodies_monet/driver.py
```diff
@@ -653,10 +652,7 @@ def mask_and_scale(self):
                             self.obj[v].data += scale
                         elif d['unit_scale_method'] == '-':
                             self.obj[v].data += -1 * scale
-                    if self.obj[v].units == 'ppv':
-                        print('changing units for {}'.format(v))
-                        self.obj[v].values *= 1e9
-                        self.obj[v].attrs['units'] = 'ppbv'
```
zmoon, line 659: I think removing this is fine, but you might want to add it to your PR description (previously ppv was automatically converted to ppbv for any model...)

melodies_monet/driver.py
```diff
@@ -1379,26 +1376,38 @@ def pair_data(self, time_interval=None):
...
                         # calculate model no2 trop. columns. M.Li
                         # to fix the "time" duplicate error
                         model_obj = mod.obj
-                        model_obj = model_obj.rename_dims({'time':'t'})
-                        model_obj = mutil.cal_model_no2columns(model_obj)
-                        #obs_dat = obs.obj.sel(time=slice(self.start_time.date(),self.end_time.date())).copy()
-
+                        i_no2_varname = [i for i,x in enumerate(obs_vars) if x == 'nitrogendioxide_tropospheric_column']
+                        if len(i_no2_varname) > 1:
+                            print('The TROPOMI NO2 variable is matched to more than one model variable.')
```
zmoon, line 1387: Maybe prefix this one with "warning" or such?

melodies_monet/util/satellite_utilities.py
```diff
@@ -151,6 +151,8 @@ def mopitt_l3_pairing(model_data,obs_data,co_ppbv_varname,global_model=True):
     # MEB: loop over time outside of regrid lowers memory usage
     for t in range(obs_data.time.size):
         obs_day = obs_data.time[t].dt.strftime(filtstr)
+        print(filtstr)
+        print(pressure_model_regrid.sel(time=obs_day))
```
zmoon, line 155: Suggest remove or adding more info about what you are printing.

melodies_monet/util/satellite_utilities.py
```diff
@@ -225,7 +227,8 @@ def space_and_time_pairing(model_data,obs_data,pair_variables):
     mod_nf,mod_nz,mod_nx,mod_ny = model_data[pair_variables[0]].shape # assumes model data is structured (time,z,lon,lat). lon/lat dimension order likely unimportant
     obs_nz = obs_data['pressure'].shape # assumes 1d pressure field in observation set
     obs_nx,obs_ny = obs_data['longitude'].shape # assumes 2d lat/lon fields in observation ser
-    
+    print(pair_variables)
+    print(mod_nz,obs_nx,obs_ny)
```
zmoon, line 231: Suggest remove or adding more info about what you are printing.

zmoon (commented):
Added a few comments

melodies_monet/driver.py
mbruckner-work: done.

melodies_monet/driver.py
mbruckner-work: Needed to apply your change while I was testing the small OMPS pairing updates due to the addition of the variable limiting for RAQMS. 

Would it be better to recommend doing this renaming through the yaml file for model output containing the tropopause pressure?

zmoon: Seems to me that it would fit well [here in the reader](https://github.com/noaa-oar-arl/monetio/blob/05b5dba9038453ec289d33cbe50ba765f85b7d9d/monetio/models/raqms.py#L164) ? So it's done automatically if ptrop is there but no error otherwise.

PR #351, opened 2025-03-25
Update ozonesonder name and fix pairing

rschwant:
This PR fixes the ozonesonder name to sonde. During this process I realized that the pairing with the latest update was not working correctly for ozonesondes anymore. The fastest fix was to just update the time to x in the tools.py vert_interp code. But then I figured I should test whether the code works correctly if it has a point in the observation above the model top. This is kind of unusual for air quality / atmospheric composition datasets that focus on the troposphere, but good for version 1 to work in all situations. It did not work correctly. So I realized that using NaN for extrapolation with pressure_model and then replacing these NaNs with the pressure_obs values for pairing does not work for all conditions. Although I did check and the update I made here does not change any current pairing for ozonesondes, FIREX-AQ, or AEROMMA (because these datasets don't get above the model top), so this update does not impact past troposphere centric work (and probably why we never noticed an issue before), but would be useful for new work that might go into the stratosphere more. 

The newest vertical pairing works like the following: 
For each point in the observations for all variables, linear interpolation and nearest extrapolation occurs for all pressure_obs available in the dataset using stratify (a very fast C++ routine). Then we need to pick the correct pressure_model point for each point in the observation file. To do this we now just have pressure_model use extrapolation linear. This gives the exact value of the pressure_obs and then selecting the right point is trivial. Not sure why I didn't think of doing this before, but this is far cleaner than the NaN approach and replacing the NaNs correctly. I think I had thought that this linear approach would not always give you the pressure_obs value, but in my tests it seems to work fine. Then I have also a NaN extrapolation that is used to print out a Note if there is a point below the lowest mid-level value in the model and a warning if there is a point above the highest mid-level value in the model. This way if someone has odd unit conversions they probably will notice right away with all the points being above or below their model values.

Let me know what you all think about these changes. I'll email 4 jupyter notebooks (I cannot attach .ipynb files and when I pdf them it is not as clear) to look though if you like for the ozonesonde and FIREX-AQ case both of which I added a fake value above the model top for testing. Again if I do not add this fake model top value, you get the same results with the old and new way. I include jupyter notebooks for this too.

### DIFF
--- a/docs/develop/development_team.rst
+++ b/docs/develop/development_team.rst
@@ -30,7 +30,7 @@ Margaret Bruckner  mbruckner-work  Docs, Satellite obs, Plots, Stats
 Rebecca Buchholz   rrbuchholz      Docs, Satellite obs, Plots, Stats
 Pablo Lichtig      blychs          Docs, Satellite and surface obs, Plots, Stats
 Quazi Ziaur Rasool quaz115         Docs, Aircraft obs, Plots, Stats
-Beiming Tang       btang1          Docs, Surface and ozone sonder obs, Plots, Stats
+Beiming Tang       btang1          Docs, Surface and sonde obs, Plots, Stats
 ================== =============== ========================================================
 
 Development Team Members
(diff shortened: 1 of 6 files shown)


colin-harkins (approved):
I tested this with your jupyter notebooks as well as on my tests I have set up for aircraft pairing and it works great! Thanks for fixing this.

PR #354, opened 2025-03-27
Fix looping error in plots, correct regression

blychs:
Fix looping order error, while maintaining capability of paired data with different observation variables.
This was a regression introduced by the wish of allowing for different pairs with different variables. It does not currently work for stats, but could probably be added doing the same.

The new PR:

- Creates obs_vars by combining all the previous observation variables, guaranteeing uniqueness and preserving order.
- It then filters the paired data to guarantee that only paired data containing the obs_vars is plotted.
- The loss of performance by doing this is negligible.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -397,8 +397,6 @@ class observation:
                         print('The variable name, {}, already exists and cannot be created with variable_summing.'.format(var_new))
                         raise ValueError
                     var_new_info = self.variable_summing[var_new]
-                    if self.variable_dict is None:
-                        self.variable_dict = {}
                     self.variable_dict[var_new] = var_new_info
                     for i,var in enumerate(var_new_info['vars']):
                         if i ==0:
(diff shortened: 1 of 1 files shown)


melodies_monet/driver.py
```diff
@@ -1628,20 +1628,27 @@ def plotting(self):
...
+                domain_infos = grp_dict.get('domain_info', {})
+                # Use only pair_labels containing obs_var
+                use_pair_labels = [p for p in pair_labels if obsvar in self.paired[p].obs_vars]
+                for domain in range(len(domain_types)):
+                    domain_type = domain_types[domain]
+                    domain_name = domain_names[domain]
+                    domain_info = domain_infos.get(domain_name, None)
+                    for p_index, p_label in enumerate(use_pair_labels):
+                        p = self.paired[p_label]
+                        obs_type = p.type
```
zmoon, line 1651: Do you want to reduce the whitespace under this to one line?

melodies_monet/driver.py
```diff
@@ -1628,20 +1628,27 @@ def plotting(self):
...
+            obs_vars = list(dict.fromkeys(obs_vars))
+
+            # loop through obs variables
+            for obsvar in obs_vars:
+                # Loop also over the domain types. So can easily create several overview and zoomed in plots.
+                domain_types = grp_dict.get('domain_type', [None])
+                domain_names = grp_dict.get('domain_name', [None])
+                domain_infos = grp_dict.get('domain_info', {})
+                # Use only pair_labels containing obs_var
+                use_pair_labels = [p for p in pair_labels if obsvar in self.paired[p].obs_vars]
```
zmoon, line 1644: How about `pair_labels_obsvar`? Current name sounds more like a Boolean variable.

zmoon (commented):
I confirm AirNow WRF-Chem works with this change.

melodies_monet/driver.py
blychs: Yes.

zmoon (approved):

rschwant:
Let's hold off on merging this yet. See my email. I'm concerned this new looping structure means that users can only pair one observation per YAML file and this was not the intent of MELODIES MONET originally. I would like to keep the flexibility that users can pair multiple obs and multiple models in one analysis / YAML script. For example, if a user wanted to compare 1 model against both AQS and AirNow data, which use the same names or OZONE and PM2.5, I don't see how this could be done with the current looping structure.

rschwant:
Or maybe this was always the case. Gotta go to a meeting, but I'll look into this further after.

blychs:
> Let's hold off on merging this yet. See my email. I'm concerned this new looping structure means that users can only pair one observation per YAML file and this was not the intent of MELODIES MONET originally. I would like to keep the flexibility that users can pair multiple obs and multiple models in one analysis / YAML script. For example, if a user wanted to compare 1 model against both AQS and AirNow data, which use the same names or OZONE and PM2.5, I don't see how this could be done with the current looping structure.

The new looping structure allows for mutlitple models and pairs, each with different variables. We can filter a bit more to allow models to not have every obs_var too. The original one did not allow for obs datasets with different obs_vars. I had changed it in the TEMPO merge and accidentally broke some other plots.

rschwant (approved):
We discussed and all looks good, thanks for the update!

PR #355, opened 2025-03-29
Hotfix: read plot defaults (reverse regression)

blychs:
This fixes a regression in the defaults of the model, that were not being read in the `timeseries` plot. I tested it with all the examples of the docs, and it seems to look exactly like what we have in `main`.
Pablo

Requested reviewers based on the meeting today.


rschwant (approved):
Looks great, thanks for the update!

PR #360, opened 2025-04-08
Update sat_l2_swath_utility.py

mlirenzhenmayi:
Corrected the amf correction for averaging kernel for TROPOMI NO2. The original code doesn't select grids.

### DIFF
--- a/melodies_monet/util/sat_l2_swath_utility.py
+++ b/melodies_monet/util/sat_l2_swath_utility.py
@@ -223,6 +223,7 @@ def cal_amf_wrfchem(scatw, wrfpreslayer, tpreslev, troppres, wrfno2layer_molec, 
     preminus         = np.zeros([nsaty, nsatx], dtype=np.float32)
     wrfpreslayer_slc = np.zeros([nsaty, nsatx], dtype=np.float32)
     tmpvalue_sat     = np.zeros([nsaty, nsatx], dtype=np.float32)
+    tmpvalue_mod     = np.zeros([nsaty, nsatx], dtype=np.float32)
     
     
     # set the surface pressure to wrf one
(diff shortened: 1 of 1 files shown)


rschwant:
@mlirenzhenmayi, I looked at this more closely and I think this works great and is more consistent to apply this correction. Should we update the examples for this update too? 

[MM_trp_no2_l2_plot.ipynb](https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/jupyter_notebooks/MM_trp_no2_l2_plot.ipynb)

[MM_trp_no2_l2_plot_wavk.ipynb](https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/jupyter_notebooks/MM_trp_no2_l2_plot_wavk.ipynb)

mlirenzhenmayi:
> @mlirenzhenmayi, I looked at this more closely and I think this works great and is more consistent to apply this correction. Should we update the examples for this update too?
> 
> [MM_trp_no2_l2_plot.ipynb](https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/jupyter_notebooks/MM_trp_no2_l2_plot.ipynb)
> 
> [MM_trp_no2_l2_plot_wavk.ipynb](https://github.com/NOAA-CSL/MELODIES-MONET/blob/main/examples/jupyter_notebooks/MM_trp_no2_l2_plot_wavk.ipynb)

Thanks Becky. I'll update these two jupyter notebooks in next PR. I also worked on submitting jobs on Hera for multiple days. And I'll commit them together as another update. Does that make sense?

rschwant:
Sure that works for me. When do you anticipate submitting the second PR? Can we get this done before version 1 release for consistency?

rschwant (approved):
Looks great! Thanks for making this correction!

mbruckner-work (approved):
Looks good to me. 
Just to clarify my understanding of what this does:
This is so that the air mass factor correction is only using model values within the tropopause (given by the TM5 output included in the satellite files) while before all grid cells in a layer would be included despite potentially being in the stratosphere if any of the cells were in the troposphere

PR #361, opened 2025-04-18
Bugfix: MOPITT vertical summing 

mbruckner-work:
While re-running some analysis with global model output I noticed that the use of `skip_na=False` in the vertical summing of the paired model data resulted in dropping of data (particularly obvious over the Rockies). 
The `skip_na` parameter had been added for pairing to regional models, thinking that it would result in effectively dropping data outside of the model domain.  However it does not take into account the situation where the satellite AK and apriori profile arrays contain nans below surface pressure.
This PR replaces `skip_na` to the more specific `min_count=1` so that if there is any valid AK, apriori, and model data for the grid cell it will remain in the column calculation.


rrbuchholz:
Good plan - glad you caught this.

rrbuchholz (approved):

blychs (approved):
I'm happy to have this merged. Nice catch, Maggie!

Note to future us: it's better to not upload changes to the jupyter notebooks themselves unless something really changed. It results in big diffs that are only due to the metadata Jupyter saves when rerun

rschwant:
This check was fixed already in this PR: https://github.com/NCAR/MELODIES-MONET/pull/363. There are no conflicts with merge, so will merge this in without the checks as the docs build error is not due to this PR and it's already fixed in develop now.

PR #366, opened 2025-05-06
update CESM figures in example

rschwant:
Update the contour plots in the CESM AirNow example on ReadTheDocs. The updates made by Pablo changed the extent of the plots which seems to impact mostly only the global coarser resolution model contour results. It probably is best to make the contour plot with all info and then zoom in, so this update seems best for these types of models. The update had no impact on regional model results with higher resolution.


blychs (approved):
It looks good to me.

rschwant:
Updated to ignore the broken NCEI link. This one doesn't seem to consistently always work, but it is the correct link for the archive, so let's ignore checking it each time.

PR #368, opened 2025-05-28
0528 Fixed Stats Update control_camchem_se.yaml

L12D2:
updated line 155 with the correct data to remain consistent with the rest of the yaml script.


rschwant (approved):
Looks great!

PR #376, opened 2025-07-04
Hotfix/docs casper env

blychs:
Fixing the Casper Machine specific instructions.
The MM default environment does not exist anymore, so it was eliminated.

### DIFF
--- a/docs/appendix/machine-specific-install.rst
+++ b/docs/appendix/machine-specific-install.rst
@@ -39,7 +39,7 @@ Otherwise, follow the :ref:`user install instructions <user-install-instructions
        $ conda activate melodies-monet
        $ conda install -c conda-forge ipykernel
 
-   Now, when opening the JupyterHub, you should be able to select the melodies-monet kernel you created.
+   Now, when opening a notebook in JupyterHub, you should be able to select the melodies-monet kernel you created.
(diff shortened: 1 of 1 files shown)


blychs:
Would @rrbuchholz and/or @lkemmons and/or @davidfillmore want to check this? I belive it should be someone from NCAR, since it's the Casper/Derecho specific instructions.

EDIT: fix typo

docs/appendix/machine-specific-install.rst
lkemmons, line 1: line 42 could be:
Now, when opening a notebook in JupyterHub, you should be able to select the melodies-monet kernel you created.
The rest looks great.

lkemmons (approved):

lkemmons:
@rschwant Can this correction to instructions for casper be pushed to the stable readthedocs page?

rschwant:
Not easily because right now our stable branch is our latest release which is v1.0. But this has been updated on our develop branch and is updated on the develop version of ReadTheDocs, so you could point new users to this. I would like to make a v1.1 in the next 1-2 months. Would you be okay with this update going into stable at that point for the v1.1 release?

lkemmons:
That's fine.

PR #380, opened 2025-07-21
Code fix to driver.py 

noribeth-m:
I added in a code fix to driver.py to be able to read in a list of multiple sorted model output files from the YAML file. 

Essentially, when reading the files from the YAML file, there are four different scenarios: 
1. If given a sorted list, assigns this list to self.files variable (change I added)
2. If the string begins with 'example', it treats the input as an ID and fetches example files. 
3. If file path uses (*), then pulls files matching the pattern using glob and sorts.
4. Checks for .txt files.

The last code change reads the list of model files given and prepares them for use in MELODIES-MONET.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -492,17 +492,20 @@ class model:
         from . import tutorial
 
         print(self.file_str)
-        if self.file_str.startswith("example:"):
+        if isinstance(self.file_str, list):
+            self.files = self.file_str # we assume the given list is already sorted
+        elif self.file_str.startswith("example:"):
             example_id = ":".join(s.strip() for s in self.file_str.split(":")[1:])
             self.files = [tutorial.fetch_example(example_id)]
         else:
             self.files = sort(glob(self.file_str))
             
         # add option to read list of files from text file
-        _, extension = os.path.splitext(self.file_str)
-        if extension.lower() == '.txt':
-            with open(self.file_str,'r') as f:
-                self.files = f.read().split()
+        if not isinstance(self.file_str, list):
+            _, extension = os.path.splitext(self.file_str)
+            if extension.lower() == '.txt':
+                with open(self.file_str,'r') as f:
+                    self.files = f.read().split()
 
         if self.file_vert_str is not None:
             self.files_vert = sort(glob(self.file_vert_str))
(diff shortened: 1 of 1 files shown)


rschwant:
Can you upload an example of how users would do this with their yaml file since it is different from what is typically done? Not sure if this is useful, but we had been linking files like the scripts here for more complex workflows too. https://github.com/NCAR/MELODIES-MONET/blob/main/examples/submit_jobs/link_files_wrf.sh

noribeth-m:
Hi. An example of the YAML file can be seen here, under models: https://github.com/noribeth-m/pair_tempo/blob/main/yaml/control_tempo_l2_no2_cesm_fv_20240701.yaml

When I added this change to my code base initially, it was because I was getting a TypeError as "TypeError: expected str, bytes or os.PathLike object, not list", when I tried to input a list to the MELODIES-MONET YAML file, whether in brackets or as is in the linked YAML file. I was specifically testing a way to pair TEMPO data with model output in a quicker way and needed to have the previous and current day as a list for the model: https://github.com/noribeth-m/pair_tempo

melodies_monet/driver.py
```diff
@@ -492,17 +492,20 @@ def glob_files(self):
         from . import tutorial
 
         print(self.file_str)
-        if self.file_str.startswith("example:"):
+        if isinstance(self.file_str, list):
+            self.files = self.file_str # we assume the given list is already sorted
```
blychs:
```suggestion
            self.files = sorted(self.file_str)
```
Is there a reason not to do this?
I am not too happy with the name file_str if it can be a list, but I guess that would require further changes to the code. Would that be doable?

melodies_monet/driver.py
```diff
@@ -492,17 +492,20 @@ def glob_files(self):
...
         # add option to read list of files from text file
-        _, extension = os.path.splitext(self.file_str)
-        if extension.lower() == '.txt':
-            with open(self.file_str,'r') as f:
-                self.files = f.read().split()
+        if not isinstance(self.file_str, list):
+            _, extension = os.path.splitext(self.file_str)
+            if extension.lower() == '.txt':
+                with open(self.file_str,'r') as f:
+                    self.files = f.read().split()
```
blychs, line 508: Is this as a guard to avoid trying to `.split()` lists? In that case, should we be able to split list elements, or is that overkill?

melodies_monet/driver.py
noribeth-m: Not at all. I think adding the sorted() would be better than inputting an already sorted list. I'll add commit this suggestion. 

As for the file_str, I could rename this in the driver. What would you suggest renaming it to?

melodies_monet/driver.py
noribeth-m: No, this snippet is meant to generate a list of strings from a text file, if that is what's provided. It would split the content of the text file by whitespaces.

blychs:
Is there anything missing? I think this could be merged into develop. @noribeth-m, can you change the target branch? We should not merge this into main until the next develop to main is done.

noribeth-m:
@blychs I just changed the target branch to develop. I don't think there is anything else missing.

blychs (approved):

blychs:
This needs approval by someone with write access. @rschwant, since you were the one looking at this before, what do you think?

rschwant (approved):
Looks great! Maggie said that this new feature worked well for her and looking at the code it keeps all the past capability working smoothly too.

PR #384, opened 2025-08-28
Adds Support for NOAA-GML NetCDF Files

shawnusaf:
Modifies driver.py and read_util.py to allow NOAA-GML files to be either in .dat or .nc formats, and adds .nc read capability along side the .dat read capability.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -171,7 +171,7 @@ class observation:
 
         _, extension = os.path.splitext(files[0])
         try:
-            if extension in {'.nc', '.ncf', '.netcdf', '.nc4'}:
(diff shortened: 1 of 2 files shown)


rschwant:
We have an example for these datasets too: https://melodies-monet.readthedocs.io/en/stable/examples/ufs-aqm-gml-ozonesonde.html

blychs:
I believe that this is more about surface sites.

shawnusaf:
Yes, the original code only reads the text version of the NOAA GML ozone files.  The pull request adds the capability to read in NetCDF versions of the NOAA GML files as well.  When specifying which GML file to use, the updated code recognizes whether it is a text file or netCDF and selects the appropriate code to read the file.

PR #385, opened 2025-09-02
Added the chimere model, and fixed a minor bug

AntonioTorga:
Added the chimere model adapter that i also pull request'ed to monetio.
Fixed a minor bug in border definition for spatial_overlay graphs. 
Might need to confirm that it was non intended behaviour but when making custom polygons for data filtering and creating a spatial_overlay graph an error occured where a float was expected but an xarray.dataarray (or dataset can't recall now) was provided.
You can check that in other if-statements around the fix the values are set to floats (like when the borders are provided for CONUS or EPA regions)
I just changed the order of the operations so it actually provides the minlat, minlon, ... for the vmodel data.

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -569,6 +569,13 @@ class model:
             print('**** Reading WRF-Chem model output...')
             self.mod_kwargs.update({'var_list' : list_input_var})
             self.obj = mio.models._wrfchem_mm.open_mfdataset(self.files,**self.mod_kwargs)
+        elif 'chimere' in self.model.lower():
+            print('**** Reading Chimere model output...')
+            self.mod_kwargs.update({
+                                    'var_list': list_input_var,
+                                    "surf_only": control_dict['model'][self.label].get('surf_only', False)
+                                    })
+            self.obj = mio.models.chimere.open_mfdataset(self.files, **self.mod_kwargs)
         elif any([mod_type in self.model.lower() for mod_type in ('ufs', 'rrfs')]):
             print('**** Reading UFS-AQM model output...')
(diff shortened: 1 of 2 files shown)


blychs (commented):
This looks good to me. 
The only reason I'm not approving it yet is that I believe that the chimere reader hasn't been merged into `monetio` yet?

melodies_monet/driver.py
```diff
@@ -569,6 +569,13 @@ def open_model_files(self, time_interval=None, control_dict=None):
             print('**** Reading WRF-Chem model output...')
             self.mod_kwargs.update({'var_list' : list_input_var})
             self.obj = mio.models._wrfchem_mm.open_mfdataset(self.files,**self.mod_kwargs)
+        elif 'chimere' in self.model.lower():
+            print('**** Reading Chimere model output...')
+            self.mod_kwargs.update({
+                                    'var_list': list_input_var,
+                                    "surf_only": control_dict['model'][self.label].get('surf_only', False)
+                                    })
+            self.obj = mio.models._chimere_mm.open_mfdataset(self.files, **self.mod_kwargs)
```
bbakernoaa: we should move away from the _modelname_mm in monetio.  These should all be exposed to the main readers

AntonioTorga: Done! fixed in monetio and the reference in driver

bbakernoaa (approved):

bbakernoaa:
Once the conflicts are resolved it looks good.

AntonioTorga:
Forgot to make it a different pull request. Going to do that in the morning so we can keep a clean git log.

PR #392, opened 2025-11-19
HERA fix to link_cartopy_files.sh Cartopy shp files now on Ursa/Hera

quaz115:
I re-download these on my end on scratch[3,4] accessible via Hera/Ursa. 

 I have downloaded them here: /scratch4/BMC/rcm1/qrasool/cartopy_shapefiles/

The source i used is: https://www.naturalearthdata.com/downloads/


rschwant (approved):
Looks great, thanks for the update!

PR #393, opened 2025-11-19
Update package list since latest netcdf4 version does not work with o…

rschwant:
Update package list in the install package to an older version of netcdf4 for now. Once the later versions work with our code, we can update this.


quaz115 (approved):
Looks good to me, already tested by Sean and working to fix the issues

rschwant:
Thanks, I also tested it on the WRF-Chem AirNow example and it fixed the problem.

PR #402, opened 2025-12-15
Delete the prefix '_' and suffix '_mm' from sat readers

blychs:
Delete the `_` prefix and the `_mm` suffix from the satellite readers. Marked as draft to wait for merging of https://github.com/noaa-oar-arl/monetio/pull/253 in `MONETIO`


rschwant:
This one needs to be updated to make it backward compatible with MONETIO

blychs:
Closed. Replaced by #421

PR #403, opened 2025-12-15
Satellite filtering

blychs:
This adds the filtering capabilities for satellite data. I. e., it just takes our existing methods and iterates them over a dict, if needed.

There are a few simple steps that would make it better for version 2, as well as more flexible:

a) We could just have all of this methods as functions in the tool. This way, our filtering, masking, etc., would work for both satellite and surface, which would help us move in a Do Not Repeat Yourself direction and make life considerably easier.

b) If we move out of dicts for satellites, we _might_ be able to call our methods directly. Seems like a long shot, though.

Cheers
Pablo

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -312,6 +312,10 @@ class observation:
         except ValueError as e:
             print('something happened opening file:', e)
             return
+        from .util.tools_sat import mask_and_scale_sat, sum_variables_sat, filter_obs_sat
+        filter_obs_sat(self)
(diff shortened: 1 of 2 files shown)


rrbuchholz:
Hi Pablo,
I think this looks OK to me. There is apparently a conflict with the driver though that needs to be fixed before merging.
Best,
Rebecca

rrbuchholz (approved):
Approved once driver conflicts resolved

rschwant:
Closing and will reopen with new driver structure

PR #404, opened 2025-12-19
Divided the driver into separate files in a Package. 

AntonioTorga:
Basically that. Separated all the clases from the driver into model, observation, pair and analysis files. Each file contains just one class.

Changed all the import statements from relative to absolute, to prevent errors later when moving files around the repository.

Changed the class names to meet the PEP8 standard.

I already tested it with a big control file, but would really like for someone else to test it with other data. Mainly because i've been having some issues with external dependencies (netCDF4 and dask).


blychs (commented):
Hi Antonio,
Thanks for doing this. I like the breaking up of the driver, which since now driver is a directory shouldn't really change much. I also like you going for absolute rather than relative imports.

My only worry is with the change of class names. It is true that PEP8 asks for CamelCase in the case of classes, but wouldn't this break our current scripts, since they all call of `an = driver.analysis()` ? If that is the case, I think we should either add the lowercase names as an alias or wait for v2 to do this, since v1.1 shouldn't have breaking changes.
Maybe you already did this, and I didn't see it.
Cheers
Pablo

AntonioTorga:
Oh yeah! You are right that would break scripts. 
I could:
1- Export them from the package with an alias.
2- Undo the changes for now.

I think the second one would be safer, can't think of a case where the first one fails but that doesn't mean it doesn't exist. So i'll add it to this pull request.

blychs:
Sounds good

AntonioTorga:
This is ready. 
Changed back the class names to lowercase and ran all the tests. Everything seems to work as intended.

melodies_monet/driver/__init__.py
```diff
@@ -0,0 +1,13 @@
+from .model import model
```
blychs:
```suggestion
from melodies_monet.model import model
```

Just to keep it consistent with the move towards absolute imports. If there is a reason to keep it relative, feel free to reject this.

melodies_monet/driver/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .observation import observation
```
blychs:
```suggestion
from melodies_monet.observation import observation
```
Same as before

melodies_monet/driver/__init__.py
```diff
@@ -0,0 +1,13 @@
...
+from .pair import pair
```
blychs:
```suggestion
from melodies_monet.pair import pair
```
same as before

blychs (commented):
Just a few (very minor) suggestions. Let me know and I'll be happy to approve it

AntonioTorga:
> Just a few (very minor) suggestions. Let me know and I'll be happy to approve it
Yeah! i think inside packages it's not that necessary but i agree that for consistency it would be better to have it as absolute.

Althought it has to be something like: 
from melodies_monet.driver.module import class
Instead of 
from melodies_monet.module import class

Already fixed and tested. Worked well.

blychs (approved):
Great job, @AntonioTorga !!!

rschwant:
Great job @AntonioTorga, do you want to resolve the conflicts? Perhaps something got merged in after you did this?

AntonioTorga:
Hi Becky! Merge conflicts are now solved and tested :)

blychs (approved):

melodies_monet/plots/surfplots.py
```diff
@@ -861,7 +861,7 @@ def make_spatial_overlay(df, vmodel, column_o=None, label_o=None, column_m=None,
     # For unstructured grid, we need a more advanced plotting code
     # Call an external function (Plot_2D)
     if vmodel.attrs.get('mio_has_unstructured_grid',False):
-        from .Plot_2D import Plot_2D
+        from melodies_monet.plot.Plot_2D import Plot_2D
```
rschwant:
```suggestion
        from melodies_monet.plots.Plot_2D import Plot_2D
```

rschwant:
I looked through the code and re-ran the doc examples.

@AntonioTorga I just found one small problem above. If we can update this, then I can re-run the AirNow CAM-Chem SE example (which is how I found this problem) and double check it now works and then I can approve and merge this in tomorrow.

AntonioTorga:
@rschwant Done! sorry for the delay. Let me know if you find any other errors.
Would be good to have more extensive checks so we can catch things like this earlier

rschwant (approved):
Great thanks, this now passes all the examples and agreed we need to make all the ReadTheDocs examples test cases too so we find problems faster and I don’t have to re-run them and manually check all the time. Hopefully we can do this soon.

PR #405, opened 2025-12-30
Update TROPOMI and bugfixes

blychs:
Add tropomi reader (which will be outsourced to MONETIO soon), and other minor bugfixes. This should be slowly merged into develop, but for now it stays in ACOM's specific branch


blychs:
@pfisterggp I don't have write access anymore, so if you could merge it to keep it working as talked, that would be great

PR #406, opened 2025-12-30
Read vmin_plot and vmax_plot

blychs:
This PR fixes the fact that vmin_plot and vmax_plot were often not read by the driver, especially if they were written in data_proc in the plots, which was inconsistent.

There is, however, one breaking change: spatial_dist was reading vmax and vmin. That was an inconsistency, so I believe that doing this breaking change is justifiable in this case.
Cheers
Pablo


blychs:
Converted to draft since there are some bugs

blychs:
I'm going to close this and redo it from scratch, since it's easier than fixing this after #404

PR #419, opened 2026-03-02
[develop] docs: add aqm-eval to forecasts

benkozi:
* Adds brief text, links, and a diagram describing the UFS-AQM MM wrapper.


docs/applications/forecasts.rst
```diff
@@ -24,3 +24,14 @@ are below.
 .. figure:: /_static/figures/OZONE_EPA_f00_rapchem.png
 
 .. figure:: /_static/figures/PM25_EPA_f00_rapchem.png
+
+UFS-AQM (AQM-Eval)
+------------------
+
+The UFS-AQM scientific team, in collaboration with NOAA-EPIC, developed a standardized MELODIES MONET model evaluation wrapper optimized for rocoto-based workflows running in HPC environments. The suite comprises packages targeting key measures of atmospheric composition model performance: chemistry, meteorology, particulate matter, and volatile organic compounds (VOCs). Automated workflow generation is currently implemented for the `UFS-Short-range Weather App (UFS-SRW) <https://ufs-srweather-app.readthedocs.io/en/develop/UsersGuide/BuildingRunningTesting/AQM.html#melodies-monet-mm-evaluation>`__ with plans to generalize workflow creation and configuration to connect to a wider variety of modeling systems.
+
+Code and wiki-based documentation are hosted on GitHub in NOAA-EPIC’s `AQM-Eval <https://github.com/NOAA-EPIC/AQM-Eval/wiki/aqm%E2%80%90mm%E2%80%90eval>`__ repository.
```
zmoon: `%E2%80%90` seems to be[^1] the unicode hyphen U+2010. Would it be possible to change the Wiki to use the standard ASCII hyphen instead?

[^1]: `from urllib.parse import unquote; f"U+{ord(unquote('%E2%80%90')):04X}"`

benkozi: I switched to hyphen. My bad on the paste error...

zmoon: I think the paste was correct, i.e. the change to ASCII hyphen is needed on the Wiki side too for this to work.

benkozi: Sorry, not sure what the ask is. I regret putting hyphens in the wiki page titles because I had forgotten how weird they are on the GH side. Can we revert to the original URL? Maybe there's a wiki setting I'm missing though.

zmoon: Sorry it's of course fine to revert if it's not possible to also change it on the GH wiki side. I was assuming it was.

benkozi: No problem. It was worth looking into again. With the hyphens left in, the wiki always want to create new pages even though the displayed browser url has them in ASCII form. 🤷

rschwant (approved):
Looks great, thanks for adding this into the docs!

PR #420, opened 2026-03-02
OH reactivity plot example in jupyter notebook for aircraft paired ne…

sean-youn:
This is an example jupyter notebook to create OH reactivity plots (e.g. for VOCs) from paired NetCDF files. This example is for paired aircraft data using RECAP-CA paired with WRF-Chem.


rschwant (commented):
Looks great thanks for adding this plot example! Here's just a couple thoughts to get this finalized.

1) Can you upload the yaml file as an example for how you would create the paired NetCDF files with this script and then include the name of it in the header for reference. 

Also we are moving toward writing which version all the yaml files have previously worked with, so when you do this, add the following to the top of the file "# This YAML file has been last tested with MELODIES MONET version 1.1."

2) Above box 8, you state that the units should be mbar for pressure. Are you sure you are not using Pa here? Typically we convert to Pa when doing the pairing so most of the time pressure_obs would be Pa and your equations below look like you are using Pa too? And can you add that you expect all VOCs to be in ppb just to keep track of all the units.

3) For the monoterpenes, great idea to use the speciation in Pfanerstill 2013 for Central Valley, but you do want to map all of these species to our 3 surrogates not just alpha-pinene and limonene, so that the rates add up to 100%. You can map no double bonds to ECLP, one double bond to API, and 2 double bonds to LIM.

4) Instead of in box 12 where you are finding the first NaN index in the obs and then using that same data, you should NaN the values in the model where the obs are NaN throughout the entire dataset. Sometimes there are NaNs throughout, not just initially. I often do this on the data pair rather than on the entire dataset so you don’t have to drop as much data (e.g., So whenever methanol obs is NaN I drop the methanol model results too, etc. for each species pair.). You should do this before you calculate the medians in box 12. 

For example, we do this in MM here in the code to show you how this works. https://github.com/NCAR/MELODIES-MONET/blob/develop/melodies_monet/driver/_analysis.py (Line 1393)

But we do this per plot, because many times with time series plots people like to not NaN all the model data and just plot the obs on top, so we give users options. When calculating statistics which is basically what you are doing on this plot, you must drop the model data when the obs are NaN for each pair, so that you are comparing your results consistently. 

Let me know if you have any questions and all else looks great and I'm excited to have this option for people to use!

sean-youn:
@rschwant Sorry I didn't quite get around to finishing this before I left. Is there still time to get these changes in, or is the code frozen for now?

rschwant:
There is still time. We are still testing all of our examples, so let's get it in.

rschwant:
Sean opened a new PR so closing this one without merging as it has been updated.

PR #421, opened 2026-03-23
Use new sat readers

blychs:
Dealing with new satellite readers.

The code looks now for the new satellite reader names, and switches them back to the old names if needed, using try/except blocks.

The code already includs #410 , let me know if I should roll that back.
Cheers
Pablo


rschwant (approved):
Looks great, thanks for the updates!

PR #425, opened 2026-04-07
Correcting developer install instructions

rrbuchholz:
Clarifying where the user would need to connect their own version of the code to contribute.


rschwant (approved):
Looks great! Thanks for the updates!

PR #429, opened 2026-04-25
Fix CI

zmoon:
* update Action versions
* add missing license headers
* update links
* fix tests
* fix docs build
  - `/home/runner/work/MELODIES-MONET/MELODIES-MONET/melodies_monet/driver/analysis.py:docstring of melodies_monet.driver.analysis.analysis:1:duplicate object description of melodies_monet.driver.analysis, other instance in api/melodies_monet.driver, use :no-index: for one of them`
  - add a missing sat rename
* update RTD


docs/environment-docs.yml
```diff
@@ -7,12 +7,12 @@ dependencies:
...
   # melodies_monet deps
-  - matplotlib-base<3.9  # for pandas v1 compat
+  - matplotlib-base
   - monet
   - monetio
   - netcdf4
   - numpy
-  - pandas<2
+  - pandas=2
   - pillow<10
```
zmoon:
```suggestion
  - pillow
```
I imagine it'd be okay to remove this upper bound now, but need to test to confirm.

c8f5f23cf added it in 2023.

rschwant: Yes, Pablo was asking about this? Has MONET and MONETIO been updated to use pandas=2 now. I have only been testing with Pandas 1 so far.

zmoon: Yep pandas 2 should be fine now in the surface data loaders we use (official in [0.3.1](https://github.com/noaa-oar-arl/monetio/releases/tag/v0.3.1)). Though of course report any issues! My comment here was about pillow. CI is already passing with pandas 2 here.

zmoon: Going to try it and see what happens with the docs build.

Edit: not that helpful since the idealized notebook doesn't show the logo anyway since debug true. @rschwant if you want to maybe create a fresh env based on the `environment-docs.yml` and see if logos look normal in figs in one of your notebooks?

rschwant: Okay I checked the WRF-Chem AirNow example, and with loading the environments via environment-docs.yml and the logos look normal using pillow = 12.2.0, so I think this works with later versions now. That's great we can use pandas=2 now too. Let's update this now and we can keep testing and if we find other things to update that doesn't work with pandas 2 we can update them then.

rschwant (commented):
Just one question: Why rename the driver files _analysis.py etc. instead of just using analysis.py?

zmoon:
> Just one question: Why rename the driver files _analysis.py etc. instead of just using analysis.py?

That was to address this issue from the API doc generation: `melodies_monet/driver/analysis.py:docstring of melodies_monet.driver.analysis.analysis:1:duplicate object description of melodies_monet.driver.analysis, other instance in api/melodies_monet.driver, use :no-index: for one of them`. Because there is a module `melodies_monet.driver.analysis` and the `analysis` class is also imported into `melodies_monet.driver` (giving a 2nd `melodies_monet.driver.analysis`).

Another way to address this would be to CamelCase the class names, but I think it was decided not to do that for now.

rschwant (approved):
Great thanks, this works for me!

docs/develop/developers_guide.rst
```diff
@@ -54,9 +54,10 @@ these instructions:
 
        $ conda create --name melodies-monet python=3.11
        $ conda activate melodies-monet
-       $ conda install -y -c conda-forge pyyaml pandas=1 "matplotlib-base<3.9" monet monetio \
-         "netcdf4<1.7" "setuptools<70" "dask>=2024.2.1" wrf-python typer rich pooch jupyterlab \
-         metpy windrose statannotations
+       $ conda install -y -c conda-forge pyyaml pandas=2 monet monetio \
```
blychs, line 57: Do we need to fix pandas=2, or can we let it choose it itself? Pandas 3 has been recently released, does it remove functionalities that affect our code? (I have not tested it)

docs/environment-docs.yml
```diff
@@ -7,13 +7,13 @@ dependencies:
...
   # melodies_monet deps
-  - matplotlib-base<3.9  # for pandas v1 compat
+  - matplotlib-base
   - monet
   - monetio
   - netcdf4
   - numpy
-  - pandas<2
-  - pillow<10
+  - pandas=2
```
blychs, line 15: Same as before, do we need to fix it to version 2? Does v3 remove functionalities we need (I have not tested it)

melodies_monet/driver/__init__.py
blychs, line 1: I just saw the discussion with @rschwant about this renaming.
How about the following?

- For now, we go with this leading "_"
- For v2, we rename to CamelCase, and add an alias with a deprecation note to avoid breaking this
- For v3, we truly deprecate the snake_case names.

If you agree, we can create an issue with that and move on.

blychs (approved):
Great! Happy to approve, just a few comments more out of curiosity than a real critique. Thanks!

melodies_monet/driver/__init__.py
zmoon: Sounds good to me @blychs

docs/develop/developers_guide.rst
zmoon: > Pandas 3 has been recently released

This was my motivation here, that it's still pretty fresh. I've made sure the monetio test suite passes with pandas 3, but there may be other parts of the ecosystem that people may be using that haven't caught up yet. And I haven't tested MM workflows with pandas **3** either.

PR #430, opened 2026-05-08
Update models to use new model reader structure in MONETIO

rschwant:
This PR updates MELODIES MONET so that the new model reader names are used instead of the old model names in MONETIO which had an _ and mm attached.

Try statements are added so that this code remains backward compatible.

The try statements for models that did not already have a reader call the model.py script first and then try the _model_mm.py script.

For models that had an outdated reader like cmaq and camx. The _model_mm.py script is called first so if this script exists it uses this one and then the model.py script is called so that if someone uses an older version of MONETIO, they do not accidentally use the old cmap.py and camx.py files in monetio that do not work perfectly with MELODIES MONET. And I added a note so people understand why this exists like this.

This is a bit clunky, so let me know if you have another suggestion for how to do this?

### DIFF
--- a/melodies_monet/driver/_model.py
+++ b/melodies_monet/driver/_model.py
@@ -177,7 +177,12 @@ class model:
                 self.mod_kwargs.update({"fname_surf": self.files_surf})
             if len(self.files) > 1:
                 self.mod_kwargs.update({"concatenate_forecasts": True})
-            self.obj = mio.models._cmaq_mm.open_mfdataset(self.files, **self.mod_kwargs)
+            try:
+                self.obj = mio.models._cmaq_mm.open_mfdataset(self.files, **self.mod_kwargs)
+                #Note this is tried first because if connect with old MONETIO code you need to use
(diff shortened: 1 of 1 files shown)


rschwant:
Need to merge this PR and then can merge this one in MONETIO: https://github.com/noaa-oar-arl/monetio/pull/277

blychs (commented):
It's looking good. I would just keep it more consistent (try always the new version before the old name, or vice versa). Right now,  CAMX and CMAQ try the old version first, and the rest try the new version first.

blychs (approved):

blychs:
> It's looking good. I would just keep it more consistent (try always the new version before the old name, or vice versa). Right now, CAMX and CMAQ try the old version first, and the rest try the new version first.

For future reference, Becky just explained this is for compatibility reason with duplicated readers in MONETIO, that are going to dissapear soon.

zmoon (approved):

PR #431, opened 2026-05-11
Update various items on the docs in prep for version 1.1

rschwant:
Still a draft. I'm still working on a couple things, but submitting now so people know what is coming and let me know what else I should add to the list.

- Update versions of MELODIES MONET, MONET and MONETIO for publications description in docs
- Update developer’s guide with new package versions so instructions work to build the docs
- Update development with latest roles from NOAA
- Update region_section from Pablo
- Update supported datasets model description for CESM SE based on developer team meeting discussion
- Update yaml appendix with the current approved model names
- Update dataset description to match new [model.py](http://model.py/) names rather than _model_mm.py names in MONETIO
- Update examples to provide a table of all the files people can use and run and instructions for how to run them based on the tutorial
- Add in a new section called “How to Begin”
- Update the Getting Started section order to connect with How to begin section
- Update developer’s guide to have clearer instructions for setting up a development environment
- Update diagram with new driver structure.
- Update read example to include limitations to make it clearer which pieces of the YAML file you can update when reading in a paired file and making plots.
-Update WRF-Chem linking script to include description on how it works so there is less confusion on how this linking is set up.


rrbuchholz (approved):
Looks good to me!
The "how to begin" section is great - I like how it identifies expectations about what we expect people to know.

rschwant:
I just updated to include yaml option descriptions for the wind barb options that I'm adding in, in a different PR. I also updated to suggest users use "cartopy=0.24" as a package to install because the latest cartopy 0.25 causes a lot 100s of annoying warnings in our examples that others have pointed out online too (https://github.com/scitools/cartopy/issues/2646). We could update the code to ignore these warnings or fix it, but it seems like this will be updated in later versions of cartopy or other packages, so seems best just to use a slightly older version while they find a more permanent fix and then we can update our code accordingly. Users can always use the latest version of cartopy and ignore the warnings too. The plots are identical.

rrbuchholz:
Yes - those small updates don't look to conflict with anything.

PR #432, opened 2026-05-18
Fix bug in looking for the new tempo reader

blychs:
The name of the new reader was wrong (it included an _no2 that is not there). I don't know how I missed it.
CHeers
Pablo


rrbuchholz (approved):
looks good!

zmoon (approved):

PR #433, opened 2026-05-19
add satellite filtering options

blychs:
Adding as draft because it still needs some testing.
This PR is a bit ugly: it uses our existing capabilities to add all of the filtering, masking and summing capabilities to satellites, by using some ugly copy-paste and looping to deal with the possibility of having dictionaries instead of a dataset. I believe this can be easily superseded when we move to Datatrees.
I included here #432 , since I think it will be merged faster and this should avoid merge conflicts.

### DIFF
--- a/melodies_monet/driver/_observation.py
+++ b/melodies_monet/driver/_observation.py
@@ -191,6 +191,11 @@ class observation:
         None
         """
         from melodies_monet.util import time_interval_subset as tsub
+        from melodies_monet.util.tools_sat import (
+            mask_and_scale_sat,
+            sum_variables_sat,
+            filter_obs_sat,
+        )
         from glob import glob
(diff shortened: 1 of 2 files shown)


rrbuchholz (approved):
Looks like a good start to me!

rrbuchholz:
Not sure why issues are happening in the CI test? Seems to maybe be related to AERONET?

dattilo (approved):
Checking approval workflow with claude enterprise transition

rschwant:
We realized that after 30 days we are unable to re-run these tests. They are failing due to NASA download issues and not due to updates from this PR, so this PR has an exception for not needing to pass these tests and in the future we will be sure to re-run these tests within the 30 day window.

PR #434, opened 2026-05-20
Fix wind barbs to work with WRF-Chem and add more options for users to adjust

rschwant:
This PR fixes the wind barbs, so they work with WRF-Chem and also adds in some additional options for users. I also added the wind barbs as a test in the WRF-Chem example script, which I had not actually tested before. Sorry for not checking that, but it works well with these updates. Let me know if I am missing anything on this too.

More specifically:
-Add options for skipping wind barbs to declutter plot
-Add kwarg options to match with the matplotlib.axes.Axes.barbs function so users can change color, length, linewidth, etc.
-Add an import error for windrose and statannotations to remind users to install these packages, if they want these features.
-For the import error for metpy, I realized we were missing a space, so that when the lines are combined it is readable.
-Change the extent crs option back to ccrs.PlateCarree() for all plots as this is most consistent and what it was originally, but I updated this in Liam's met PR to be consistent with the windbarbs and now realize that was not quite right. 
-Then for the windbarbs, we should always be using ccrs.PlateCarre for the transform since all of our lat long data is on degrees for model data and paired data, so updated this too.
-For models that do not use dask for reading in the data like wrf-python for now, the dataarray for the spatial_overlay plots for the windbarbs was erroring, so I change this to add ".values" so that the extra metadata is not provided and cartopy does not get tripped up with this.
-The example I uploaded prior for WRF-Chem did not test turning on the wind barbs, which is why I didn’t notice this before. So I updated the example for this and tested it with all the same features as the others and this is now uploaded too and working well.
-I then re-ran all the wind barb examples in the docs to make sure everything works well now for the other models too and to describe the new features.
-I updated wrf example yaml file with the note that it was tested with MELODIES MONET v1.1 that we are now moving toward.
-Update the skipping on the spatial_overlay plot so that skip in both x and y directions, so that regional model wind barbs look cleaner too and not just the 1 deg models.

Let me know if there are any other updates people think would be useful for these wind barbs too.

### DIFF
--- a/docs/examples/control_ish_ufschem-example.yaml
+++ b/docs/examples/control_ish_ufschem-example.yaml
@@ -234,7 +234,7 @@ plots:
   plot_grp3:
     type: "spatial_overlay"
     fig_kwargs:
-      states: True  # such as whether to show the state boundaries
+      states: True  # whether to show the state boundaries
       figsize: [10, 5]
     text_kwargs:
       fontsize: 16. 
--- a/docs/examples/control_ish_ufschem-example.yaml
+++ b/docs/examples/control_ish_ufschem-example.yaml
@@ -234,7 +234,7 @@ plots:
   plot_grp3:
     type: "spatial_overlay"
     fig_kwargs:
-      states: True
+      states: True  # such as whether to show the state boundaries
       figsize: [10, 5]
     text_kwargs:
       fontsize: 16. 
@@ -245,6 +245,7 @@ plots:
       rem_obs_nan: True
       set_axis: True
       wind_barb: True
+      wind_barb_skip: 3
       filter_string: '24 < latitude < 50 and -130 < longitude < -60'
 
   plot_grp3a:
@@ -261,6 +262,7 @@ plots:
       rem_obs_nan: True
       set_axis: True
       wind_barb: True
+      wind_barb_skip: 3
       filter_string: '24 < latitude < 50 and -130 < longitude < -60 and ws != 0 and wdir != 0' #Remove observations labeled as calm from the analysis
 
   plot_grp3b:
@@ -277,6 +279,11 @@ plots:
       rem_obs_nan: True
       set_axis: True
       wind_barb: True
+      wind_barb_skip: 3
+      wind_barb_kwargs:
+        length: 6
+        linewidth: 0.85
+        color: "blue"
       filter_string: '37 < latitude < 49 and -104 < longitude < -80'
 
   plot_grp4:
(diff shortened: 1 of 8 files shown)


melodies_monet/util/metcalc.py
zmoon, line 1: don't think these changes are necessary

docs/examples/control_ish_ufschem-example.yaml
```diff
@@ -234,7 +234,7 @@ plots:
   plot_grp3:
     type: "spatial_overlay"
     fig_kwargs:
-      states: True
+      states: True  # such as whether to show the state boundaries
```
zmoon:
```suggestion
      states: True  # whether to show the state boundaries
```

melodies_monet/plots/surfplots.py
```diff
@@ -299,6 +300,10 @@ def make_spatial_bias(df, df_reg=None, column_o=None, label_o=None, column_m=Non
         Name of v_component in the model to use for wind barbs
     wind_barb : boolean
         Whether to plot wind barbs (True) or not (False)
+    wind_barb_skip : integer
+        Frequency to skip wind barbs
```
zmoon: To me this seems a confusing way to describe it. Maybe something with "stride" or "step" and for that matter maybe the arg should be `wind_barb_stride` instead. When `wind_barb_skip=1` (current default), you're not *skipping* 1.

rschwant: That's a good point. Which do you prefer? I can change this easily. Step may be more intuitive I guess, given that python calls it a step typically. https://www.w3schools.com/python/numpy/numpy_array_slicing.asp

melodies_monet/util/metcalc.py
rschwant: If you don't add the space when it combines to create the error in jupyter notebook at least it puts them together, so it reads "Install with: conda install -c conda-forge **metpyor** with: pip install metpy" instead of "Install with: conda install -c conda-forge **metpy or** with: pip install metpy"

melodies_monet/plots/surfplots.py
zmoon: step works for me

melodies_monet/util/metcalc.py
zmoon: 🤦‍♂️ oh yes sorry I was reading it wrong

melodies_monet/plots/surfplots.py
rschwant: This is now updated to use wind_barb_step.

zmoon (approved):

PR #435, opened 2026-05-21
BUG: Ensure plot_dict is a dict in plot_timeseries when not given

DWesl:
Allow the function to succeed even when the optional argument is not given.

### DIFF
--- a/melodies_monet/plots/surfplots.py
+++ b/melodies_monet/plots/surfplots.py
@@ -556,6 +556,8 @@ def make_diurnal_cycle(df, column=None, label=None, ax=None, avg_window=None, yl
         text_kwargs = {**def_text, **text_dict}
     else:
         text_kwargs = def_text
+    if plot_dict is None:
+        plot_dict = {}
     # set ylabel to column if not specified.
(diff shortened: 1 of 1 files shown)


zmoon (commented):
It looks like some other functions in this module would benefit from this treatment too, if you don't mind

DWesl:
I added the ones I saw.  I could have `mypy` or similar look for more if I added type annotations, or found a type-checker that understands NumPy-style docstrings (didn't find that, but scientific-python's docstub utility can use docstrings to make type-stub files, which serve the same purpose).

rschwant:
Thanks for the fix! Can you update this PR to go to our develop branch instead of the main branch? We only take updates to the develop branch from the community.

DWesl:
Not with my current permissions, and possibly not at all.  I can, however, reproduce these changes on top of develop and submit a new PR

rschwant:
Yes, do you mind resubmitting the changes to the develop branch? We can just close this one. That's probably the easiest method.

DWesl:
Replacement submitted as #438

PR #436, opened 2026-05-21
ENH: Adapt model rename_vars for cfgrib conventions

DWesl:
cfgrib names the valid time variable `valid_time`, where this seems to expect the pynio default, which seems to be `time`.  This sets the variable name to conform to expectations.

(cfgrib also has a "time" variable, used for the forecast reference/initialization time)

### DIFF
--- a/melodies_monet/driver.py
+++ b/melodies_monet/driver.py
@@ -652,6 +652,8 @@ class model:
                     if 'rename' in d:
                         self.obj = self.obj.rename({v:d['rename']})
                         self.variable_dict[d['rename']] = self.variable_dict.pop(v)
+        if "valid_time" in self.obj.dims:
+            self.obj = self.obj.rename(valid_time="time")
 
     def mask_and_scale(self):
         """Mask and scale model data including unit conversions.


bbakernoaa:
Just to note that we should avoid using cfgrib as it can’t be used on NOAA
operational machines at all. If you notice in MONETIO (v1 branch) we
purposefully didn’t use cfgrib but used grib2io developed within noaa that
correctly supports composition data without the need of local grib tables
(as is done in cfgrib)

I also wonder why we have IO happening in melodies-monet. If the decision
is to bypass monet and monetio all together that’s fine I just want to know
as it’s holding up some development on that side.

DWesl:
Was not aware of grib2io as an XArray backend, and didn't see that on [the melodies-monet dependencies list](https://melodies-monet.readthedocs.io/en/stable/getting_started/installation.html), likely because it's properly a dependency of [monetio](https://monetio.readthedocs.io/en/stable/installing.html).  The only grib engine mentioned in the MonetIO code is pynio, in `models/ncep_grib.py`.

`cfgrib` is the backend mentioned on [the XArray page](https://docs.xarray.dev/en/stable/user-guide/io.html#grib-format-via-cfgrib) for grib, so that's what I've been using.

Should I open a documentation PR, here ~~or on MonetIO~~, suggesting `grib2io` for Grib I/O?  (and perhaps on XArray, for the same purpose?)

As for the location, in part it was due to `rename_variables` only renaming data variables, not coordinates, and in part due to me forgetting about the `ncep_grib` model over in MonetIO.  I'll change model and see what happens.

EDIT: Found [the MonetIO dependency mention](https://monetio.readthedocs.io/en/v1/installing/#optional-dependencies)

bbakernoaa:
https://monetio.readthedocs.io/en/v1/

v1 has a lot of advances that hasn’t been incorporated into melodies-monet. 

Universal api loading, fully lazy loading support for UGRID and CF conventions, all data returned as xarray objects, expanded observation datasets and models and satellites. 

Monet has also undergone a major change simplifying and supporting the new data structures from monetio and leveraging xregrid (https://noaa-emc.github.io/xregrid) and delegating out statistics to the new monetio-stats repo. 

In general I think that we should converge on more widely used standards like CF or UGRID in general rather than independent libraries for specific datasets etc.

bbakernoaa:
https://github.com/noaa-oar-arl/monetio/blob/v1/monetio/readers/gfs.py#L55

PR #438, opened 2026-05-27
BUG: Ensure plot_dict is a dictionary before use

DWesl:
plot_dict is an optional argument in many functions in this file, and is frequently assumed to be a dictionary without checking whether the optional argument was provided.  This ensures plot_dict is a dictionary before use.

Replaces #435


rschwant (approved):
Looks good! Thanks for the comment on the rose plot we can add that capability to connect with the fig dictionary at some point too.

PR #443, opened 2026-06-23
Update examples for surface and aircraft and fix any bugs that come up

rschwant:
I'm re-running all the examples for surface and aircraft datasets. I updated the yaml file to state they have been tested with MELODIES-MONET, finding all the data and putting them in a consistent space, and fix any bugs that come up in the process. 

These updates include:
-Cleaned up all the examples for the surface and aircraft results. 
-For the aircraft examples, moved FIREX-AQ examples to AEROMMA datasets as this will be easier to describe in the paper to use a more recent campaign. All examples are now stored on Meiyu or Hera in one location, so I can easily run these later on.
-Cleaned up a couple duplicative files or files that are outdated and are replaced by more complex examples as we do not need to maintain both options
-Added a note for all YAML files I tested and updated that they work with MELODIES MONET version 1.1
-Fixed violin plot y_labels to default to the obs name if no label provided
-Added a note to the curtain plot YAML file example that the plot only plots all model data and obs data regardless of how you set rem_obs_nan currently, but we will add more options in the future.
-The secondary y-axis for altitude was sometimes not working properly based on you set the ylimits and intervals. So switched to plotting with pandas since the ax1 uses pandas and this was causing an inconsistency sometimes with then using matplotlib. And made the ylimits and intervals optional so users can also just let matplotlib determine the scaling or specify desired behavior if they don’t like the default. I also added that the secondary y-axis would average like the ax1 data for consistency too. This now works for all cases that I tested. Updated all examples for this including the ones on the docs.
-Copied new curtain plot from AEROMMA to replace the old FIREX-AQ plot in the docs for consistency with these examples.
-Talked to the aerosol observationalists with the AEROMMA data and they agreed we can just use the standard temperature and pressure of 273.15 K and 1013.25 mb/hPa for all of their data. So delete the AMS and BC data versions of convert_std_to_amb for consistency and updated the example to just use the generic function convert_std_to_amb instead. And update the standard temperature to 273.15 K in this function. This is much easier to explain to the community and the observationalists approve of this approach.
-Delete redundant save and read yaml file and jupyter notebooks which is already in the doc examples.
-Updated the SUNVEx ground example to pair with O3, NO, and NO2 because current example was paring PM2.5 with Org aerosol so that everything in the example now is scientifically correct.
-Update fire mobile case to be SUNVEx for one of the mobile drives for consistency and hard to find all these older datasets.
-The aircraft AEROMMA UFS-AQM pairing failed with Pandas 2 due to a float32 and float64 mismatch in the pairing variables for vertical interpolation. Added for latitude, longitude, and pressure that you upcast to float64 for the model data to have consistency in the observations.
-Also added to reset_index() for the df_obs dataframe to have time be a column like the df_model dataframe before applying merge_asof. This does not cause problems yet, but not sure Pandas will always remain so flexible and it's best practice when you are merging that the the time variable is a column in both dataframes rather than one as an index and one as a dataframe. I realized this after printing out the info identifying the floating issue.
-The AEROMMA UFS-AQM example for pairing in jupyter notebook has been moved to meiyu and all tests now work with Pandas 2.
-The AEROMMA UFS-AQM and AirNow and UFS-AQM submit scripts have been updated and tested on Hera and all works well.
-The AEROMMA UFS-AQM docs example has been re-run and uploaded but no large changes.
-The UWy sonde and UFS-Chem example has been re-run and uploaded for both pairing and plots. Pandas 2 slows down the pairing significantly and ran into memory issues using 2s versus 10s. The same updated code here works fine for 2s with Pandas 1. So if people are struggling with slower aircraft pairing code we can suggest to move to Pandas1 and look into this more later on.
-Re-uploaded WRF-Chem AirNow and WRF-Chem AirNow Reg, and save doc examples because Pandas2 reordered time which changed the paired printout, but not the results at all, but so that the paired printout is the same for people to compare with I uploaded a clean example.
-Realized the UFS-Chem examples with ISH, ISH-LITE, and AirNow were pulling in a WRF-Chem stat csv example erroneously, so deleted this and re-uploaded these too. No other changes were needed for Pandas 2 and results are the same.
-All the Docs examples work with Pandas 2 and anything I didn't update did not change. Except the CESM SE plot_d2 function does not work with Pandas 2 and I created an issue for this and we can fix it in a separate PR: https://github.com/NCAR/MELODIES-MONET/issues/449
-Small fix was needed for curtain plot moving to Pandas2 to send just the value for the min and max into np.linespace instead of the whole array which caused a dimension issue in Pandas2 but worked fine in Pandas1. I also added "constrained_layout=True" because in Pandas2 the color bar was overlapping the plots and this makes sure there is enough room for the colorbar using both Pandas1 and Pandas2.
-All examples not on docs for surface and aircraft are also retested with Pandas2 and work well.


rschwant:
Quazi and Colin (@quaz115 @colin-harkins), can you review this PR? I've updated and cleaned up all the examples and made some small updates to aircraft plots and pairing to get everything working with Pandas 2 and a couple plots like the secondary axis plot were not consistently working, so I made this code work in all cases. Let me know if you see any problems or suggest any updates. All the examples run on meiyu and the submit scripts all run on Hera. We will move these to standard doc examples soon and these are good science checks as we proceed forward as well.

melodies_monet/util/tools.py
```diff
@@ -298,6 +298,14 @@ def resample_stratify(da, levels, vertical, axis=1,interpolation='linear',extrap
...
     from pandas import merge_asof
 
+    #The model data will occationally be pulled in as float32 to save memory especially with Pandas2.
+    #While Pandas read_csv and float defaults to float64, which is used to read in the observations.
+    #Let's upcast the matching model columns to float64 to ensure consistency with observations
+    #during this pairing step, while keeping model data at float32 for other variables to save memory.
+    match_cols = ['latitude', 'longitude', 'pressure_model']
+    for col in match_cols:
+        ds_model[col] = ds_model[col].astype('float64')
+
```
quaz115, line 308: @rschwant The dtype normalization currently seems to cover only the model. Supported aircraft/sonde observations can also come from NetCDF, where latitude, longitude, and pressure_obs may remain float32. After upcasting the model, merge_asof can still fail because its by columns—the columns used to match rows—must have identical dtypes.
Could we normalize latitude, longitude, and pressure_obs on both df_obs and df_model immediately before merge_asof, after renaming pressure_model? Adding a regression test with float32 observation keys and float64 model keys would help confirm this works for both CSV and NetCDF observations.

melodies_monet/util/tools.py
```diff
@@ -514,41 +523,6 @@ def convert_std_to_amb(
...
 
 
-def convert_std_to_amb_ams(ds, convert_vars=None, temp_var=None, pres_var=None):
-    """ 
-    Backwards compatable wrapper for AMS Dataset.
-    This uses international std atmosphere defination
-    Pressure = 101325 Pa
-    Temperature = 273 K
-    """
-    return convert_std_to_amb(
```
quaz115, line 524: @rschwant  Could you confirm whether removing convert_std_to_amb_ams and convert_std_to_amb_bc is intended to break existing user imports? These functions are available in the current 1.0 release, and develop describes them as backward-compatible wrappers. A wrapper is an old function name that forwards the call to the newer generic function, allowing existing code to continue working.
Would it make sense to retain both wrappers for a deprecation period, have them call convert_std_to_amb, and issue a DeprecationWarning? This warning tells users to update their code without causing it to fail immediately. The AEROMMA examples could still use the generic function and its new 273.15 K default, while the older names could be removed in a future major release.

quaz115:
@rschwant  I found a few minor documentation and example consistency items that I wanted to clarify:
docs/getting_started/how_to_run.rst still uses the deleted examples/yaml/control_cmaq.yaml.
docs/users_guide/supported_diagnostics.rst still points to the deleted examples/yaml/control_rrfs_cmaq_airnow_reg.yaml.
The submission section in AEROMMA_UFS-AQM_Aircraft_Pairing.ipynb references the deleted run_aircraft_pairing_loop.py and a submit_hera.job file that does not appear to exist. Should these be run_aircraft_pairing_loop_AEROMMA.py and submit_hera_aircraft_AEROMMA.sh?
The changed docs controls control_looping_UWY_UFS_CHEM.yaml, control_read_looped_aircraft_AEROMMA_UFS_AQM.yaml, and control_read_looped_aircraft_UWY_UFS_CHEM.yaml do not contain the “tested with MELODIES-MONET 1.1” note mentioned in the PR description.
The package in this branch still reports melodies_monet.__version__ == "1.0".
Could you confirm the intended replacements for the deleted files and whether the 1.1 note is intended to apply only to files under examples/yaml? If 1.1 is the planned release rather than the version used during testing, identifying the tested commit or saying “tested for the planned 1.1 release” might make the validation clearer.
This is a minor consistency note and is not intended to be blocking.

quaz115:
@rschwant For reproducibility, would it be helpful to add a short validation note to the PR description or the relevant example documentation? It could identify the Pandas version, MELODIES-MONET commit, commands or batch scripts, and examples used for the Meiyu/Hera testing?
The committed notebooks do not contain error outputs, but CI currently executes only the idealized documentation notebook rather than all the surface, aircraft, and sonde workflows described here. Recording the tested environment would help future contributors reproduce these checks.

quaz115 (commented):
@rschwant Thanks for the extensive surface and aircraft example refresh. I reviewed the changed code and relevant consumers elsewhere in the branch. The plotting updates, SUNVEx mappings, AEROMMA example changes, index reset, STP update, and curtain fixes generally look consistent.
I am requesting changes for the two main concerns noted inline:
1. The pairing dtype normalization should cover both the observation and model matching columns so supported float32 observations do not still fail in merge_asof.
2. The existing AMS and BC conversion function names should have a compatibility/deprecation path so current 1.0 user imports do not fail immediately.
The remaining documentation and example consistency items are minor and non-blocking. Once the two main questions are addressed, I would be happy to approve the updates.

PR #444, opened 2026-06-24
removing old yamls

rrbuchholz:
Removing satellite yamls that don't have the header definition of where they work.


rschwant (approved):
Approved thanks for the cleanup!

rschwant:
This doc test failing is being fixed by Zach here (https://github.com/NCAR/MELODIES-MONET/pull/447) so this PR is not causing the problem.

PR #451, opened 2026-07-21
add montage() to help consolidate output into single .pngs. Esp usefu…

L12D2:
…l for vert profiles etc.

User specified optional YAML option to aggregate all MM output onto single images. Warning is fired that these montages are not meant for official publication purposes and solely are to support rapid analyses.


L12D2:

L12D2:
This is NOT intended for v1.1

PR #452, opened 2026-07-21
Regulatory fixes

L12D2:
Addressed a prior issue where regulatory information was being called in surfplots. This now exists in a regulatory util file. Need to check that all references to the regulatory surfplots are switched over to the regulatory util. If it isnt in this PR, it will be ammended in a following PR(s).


L12D2:

L12D2:
This is NOT intended for v1.1

PR #453, opened 2026-07-21
util fixes: lon-wrap region select, dim-aware paired concat, log-line…

L12D2:
This may be user error... However, Model meshes tend to be 0-360 deg. The metro domain boxes tend to be -180 to 180. So, 
“Create_autoregion” will wrap to both conventions. I had been running into plots where it would crop to the metro area, but not have any data (even though the full conus was filled).

### DIFF
--- a/melodies_monet/util/region_select.py
+++ b/melodies_monet/util/region_select.py
@@ -240,17 +240,31 @@ def create_autoregion(data, domain_type, domain_name, domain_info=None):
             f"{domain_type!r}. If you need more capabilities, check out the custom: "
             "regions capabilities. Be aware that they require regionmask."
         )
+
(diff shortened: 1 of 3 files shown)


L12D2:

L12D2:
This is NOT intended for v1.1

L12D2:
https://ncar.github.io/gdex-examples/hadisst-elnino/ 

this notebook provides some ideas of perhaps how to deal with the 0-360 degree convention.

melodies_monet/util/region_select.py
bbakernoaa, line 1: there are utilities in monet already for wrapping

melodies_monet/util/tools.py
bbakernoaa, line 1: not sure why we are not using the monet vertical interpolation here especially since it already uses stratify

PR #454, opened 2026-07-28
Add unstructured mesh support and model-agnostic regridding

L12D2:
# Unstructured mesh support and model-agnostic regridding

Foundation for evaluating unstructured-mesh models (CESM-SE / MUSICAv0, MPAS) against gridded and satellite observations. This PR is purely additive — no existing code path changes behavior.

### `util/uxarray_util.py` (new)

Mesh layer built on `uxarray`:

- `open_uxgrid` / `clean_uxgrid_from_scrip` — build a `uxarray` Grid from SCRIP, EXODUS, UGRID, or a native MPAS mesh. SCRIP files pad every cell to `grid_corners` by repeating corners; those degenerate corners make ESMF refuse to build conservative weights, so each face is depadded to its true polygon (4/6/8/10-gon for ne0CONUS dual cells). Results are cached per path — depadding a 174k-face mesh takes seconds.

- `uxgrid_from_corner_bounds` — build a mesh from satellite pixel corner bounds, so an L2 swath can act as a conservative regrid source or target.

- `subset_mesh_to_bbox` / `subset_model_source` — crop mesh + data to a bounding box before building weights. Building ESMF weights against a full 174k-face mesh makes conservative regridding OOM (extremely heavy); cropping to the swath footprint helps fixes it to some extent. We really need parallelization here... 

- `flatten_to_faces` / `faces_to_grid` — stack 2-D swath dims to a flat face dim and back.

- `sample_unstructured_at_points`, `to_neg180` — point sampling, and 0–360 ↔ −180..180 normalization (without it a CONUS model and a CONUS swath compare as disjoint).

---

### `util/regrid_util.py`

Adds a model-agnostic `regrid()` alongside the existing xESMF config path (`setup_regridder` / `filename_regrid` are untouched).

Dispatches on method:

#### `ckdtree`
- `nearest_s2d`
- `nearest_d2s`
- `radius_mean` 

Fast point sampling via a cached `cKDTree`; **does not conserve mass.**

#### `xregrid` / ESMF
- `conservative`
- `bilinear`
- `patch`

True area-weighted regridding.

Also includes:
- a coverage-normalization fix: the installed `xregrid`'s conservative weights row-sum to ~2 rather than 1, which silently doubles every conservative product. `_coverage_normalize` corrects this.
- Dask-parallel weight generation is env-gated (`MM_XREGRID_PARALLEL`, default off) behind a `UxDataset.isel` slice shim.
- ESMF meshes are explicitly released per regridder — without that, a multi-day pairing loop leaks hundreds of MB per granule.

---

### `plots/uxarray_render.py` (new)

- `render_unstructured_field()` renders a field on its native mesh, subsetting before building polygons so a zoomed view doesn't construct every face.

---

### `driver/_model.py`

Unstructured load path for `cesm_se` and `mpas`:

- resolves the grid file
- binds `model.uxgrid`
- sets:
  - `mio_has_unstructured_grid`
  - `mio_scrip_file`
  - `mio_mesh_file`

## Dependencies

`uxarray` and `xregrid` are optional, imported with the same `try/except ImportError` guard this codebase already uses for `windrose` in `surfplots.py`. `import melodies_monet` works without them and existing runs are unaffected.

## monetio compatibility

Prefers `monetio.models._cam_unstructured` (noaa-oar-arl/monetio#286) and falls back to the existing `cesm_se` / `_cesm_se_mm` readers, so this works against released `monetio` today. MPAS `mesh_file` input requires the newer reader and raises an explicit message saying so.

## Notes

- `plots/Plot_2D.py` is intentionally left in place but is being removed in downstream PRs— `surfplots.py` still imports it for the unstructured branch. The swap to `render_unstructured_field` comes with the surface-plot updates in a follow-up PR.
- Reading `mesh_file` from the YAML isn't wired yet, so MPAS support is dormant here; CESM-SE behaves exactly as before.

### DIFF
--- a/melodies_monet/driver/_model.py
+++ b/melodies_monet/driver/_model.py
@@ -5,6 +5,11 @@ import warnings
 import xarray as xr
 import monetio as mio
 
+try:
+    import uxarray as ux
+except ImportError:  # optional dep
+    ux = None
+
 
 class model:
     """The model class.
@@ -16,6 +21,7 @@ class model:
         """Initialize a :class:`model` object."""
         self.model = None
         self.is_global = False
+        self.is_track = False
         self.radius_of_influence = None
         self.mod_kwargs = {}
         self.file_str = None
@@ -35,6 +41,10 @@ class model:
         self.plot_kwargs = None
         self.proj = None
         self.mod_to_overpass = False
+
+        # Build in uxarray native support ; populate for the scrip file provided in cesm-se 
+        self.uxds = None
+        self.uxgrid = None
 
     def __repr__(self):
         return (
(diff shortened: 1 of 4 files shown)


L12D2:

This is NOT intended for v1.1

PR #455, opened 2026-07-28
B3a tropomi swath

L12D2:
# Generic TROPOMI L2 conservative swath pairing

Adds TROPOMI NO₂ / HCHO / CO pairing against unstructured and gridded models, alongside the existing TEMPO-specific path. Append-only — no existing function changes.

## Overview

Pairs in either direction:

- Conservatively regridding the swath onto the model mesh.
- Conservatively regridding the model onto native swath pixels.

Source geometry is built from the L2 pixel corner bounds, so ESMF weights are true area overlaps rather than nearest-neighbour approximations. Weights are rebuilt per granule since the swath moves.

Includes the crop-to-extent and mesh-subset paths needed to keep memory bounded on full-mesh runs.

## Dependencies

Builds on the mesh/regrid primitives from **#454**:

- `regrid_util.regrid`
- `uxarray_util.open_uxgrid`
- `uxarray_util.uxgrid_from_corner_bounds`

All are imported lazily inside functions.

> **Depends on #454** — please merge that first.

### DIFF
--- a/melodies_monet/driver/_model.py
+++ b/melodies_monet/driver/_model.py
@@ -5,6 +5,11 @@ import warnings
 import xarray as xr
 import monetio as mio
 
+try:
+    import uxarray as ux
+except ImportError:  # optional dep
+    ux = None
+
 
 class model:
     """The model class.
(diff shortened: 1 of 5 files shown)


L12D2:

This is NOT intended for v1.1

PR #456, opened 2026-07-28
B3b tempo conservative

L12D2:
# TEMPO L2 conservative pairing rework

Reworks TEMPO NO₂/HCHO pairing to support true conservative regridding in both directions, alongside the existing nearest-neighbour path.

## What's changed

- `_conservative_mod2swath` / `_conservative_swath2mod_scan` — area-weighted model↔swath regridding, with swath geometry built from L2 pixel corner bounds and weights rebuilt per granule.

- `_apply_scattering_weights` (+ hydrostatic variant) — applies retrieval scattering weights / AMF so the model column is directly comparable to the retrieval. The four `apply_weights_mod2tempo_*` entry points are preserved as thin wrappers over this shared core.

- `_unstructured_back_to_modgrid` — returns paired swath data to an unstructured model mesh.

- Adds native-swath targets, `save_vars`, and `crop_extent` for bounded-memory runs.

- `calc_partialcolumn` is now imported from `util.tools` rather than duplicated here; the module-level import keeps `sat_l2_swath_utility_tempo.calc_partialcolumn` resolving for any external caller.

## Compatibility

No existing public function is removed — all functions on `develop` remain available.

## Dependencies

> **Depends on #454 and #455** — please merge those first.

### DIFF
--- a/melodies_monet/driver/_model.py
+++ b/melodies_monet/driver/_model.py
@@ -5,6 +5,11 @@ import warnings
 import xarray as xr
 import monetio as mio
 
+try:
+    import uxarray as ux
+except ImportError:  # optional dep
+    ux = None
+
 
 class model:
     """The model class.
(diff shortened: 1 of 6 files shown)


L12D2:

This is NOT intended for v1.1

PR #458, opened 2026-07-28
B3d tropomi routing

L12D2:
# TROPOMI L2 reader routing for NO₂, HCHO and CO

Routes TROPOMI products to `monetio`'s generic `tropomi_l2.open_datasets` reader, which returns per-granule datasets carrying the pixel corner bounds that the conservative/unstructured pairing path needs.

## What's changed

- `tropomi_l2_no2` — now branches on `sat_method`: `replace_apriori` keeps the legacy NO₂-specific reader (`preslev`/`troppres`); everything else uses the generic reader.

- `tropomi_l2_hcho`, `tropomi_l2_co` — new branches, both using the generic reader.

- The separate `sat_method == "apply_ak"` branch is folded into the generic path, since that reader is now the default for all TROPOMI products.

## Note on #433

This work was developed on a branch that predates **#433** (`satellite_filtering`, @blychs), so the extracted file was missing the `tools_sat` import and the `mask_and_scale_sat`, `sum_variables_sat`, and `filter_obs_sat` calls.

I've restored them here—the diff should contain **no #433 changes** at all, and the TROPOMI routing is orthogonal to that feature.

## Dependencies

> **Depends on #454.**

### DIFF
--- a/melodies_monet/driver/_model.py
+++ b/melodies_monet/driver/_model.py
@@ -5,6 +5,11 @@ import warnings
 import xarray as xr
 import monetio as mio
 
+try:
+    import uxarray as ux
+except ImportError:  # optional dep
+    ux = None
+
 
 class model:
     """The model class.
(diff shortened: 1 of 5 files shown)


L12D2:

This is NOT intended for v1.1

PR #460, opened 2026-07-29
Support altitude as curtain plot vertical coordinate; guard missing a…

L12D2:
Support altitude as curtain plot vertical coordinate; guard missing altitude

# Aircraft plot fixes

## What's changed

- `make_curtain_plot` gains `vert_coord` (`"pressure"` default, or `"altitude"`). With altitude the y-axis is no longer inverted, the label switches to altitude units (`altitude_units`), and tick generation handles either axis direction.

- `make_vertprofile` drops rows with no altitude before binning. Previously a flight missing the altitude variable produced NaN bins and crashed on the `.mid` lookup; now it warns and skips. Applied to both the new-axis and existing-axis paths.

- `make_spatial_bias` closes its figure after saving. My long multi-flight jobs were accumulating open figures and probably contributing to some slowdown there. 

## Dependencies

> Independent — no dependency on other PRs in this series.

### DIFF
--- a/melodies_monet/plots/aircraftplots.py
+++ b/melodies_monet/plots/aircraftplots.py
@@ -176,6 +176,8 @@ def make_spatial_bias(df, df_reg=None, column_o=None, label_o=None, column_m=Non
     
     #plt.tight_layout(pad=0)
     savefig(outname + '.png', loc=4, logo_height=120)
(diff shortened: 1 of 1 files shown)


L12D2:

This is NOT intended for v1.1

PR #461, opened 2026-07-29
B5b xarray plots

L12D2:
# Mesh-aware spatial and diurnal plots

Makes the xarray plotting path work with unstructured paired output (CESM-SE / MPAS returned on `n_face`/`ncol`) rather than assuming structured (`y`, `x`) swath shapes, and fixes several diurnal-cycle issues. No functions added or removed — all ten existing functions are preserved.

## Spatial dimension handling

`make_timeseries` reduced over a hardcoded `dim=("y", "x")`, which only holds for structured swath output. It now reduces over whatever non-time dims are present:

```python
spatial_dims = [d for d in dset[varname].dims if d != "time"]
```

Applied to all four reduction sites (raw and `avg_window`-resampled).

`make_diurnal_cycle` gets the same treatment so only time remains before hour grouping.

## Unstructured rendering

`make_spatial_dist` and `make_spatial_bias_gridded` previously always used `pcolormesh`, which requires a regular grid. They now branch:

- Structured grids keep `pcolormesh`.
- Unstructured fields render as native mesh polygons via `uxarray_render.render_unstructured_field`.

The time-mean guard also changed from:

```python
if len(var2plot.dims) == 3
```

to checking for a time dimension followed by `.squeeze()`:

```python
if "time" in var2plot.dims
```

The dimension-count test misfired on unstructured shapes, where a `(time, n_face)` field has only two dimensions.

## Domain zoom

For `auto-region:` and `custom:` domains, the map extent is now derived from where the field actually has finite data instead of the full grid bounds.

Previously a CONUS or global mesh subset to a metro box still rendered worldwide, leaving the region of interest as a few invisible pixels.

## Diurnal cycle fixes

- Fractional hour offsets now work: `np.timedelta64(time_offset, "h")` truncates to whole hours and silently discards e.g. a 5.5-hour offset, so this uses `pd.to_timedelta(time_offset, unit="h")`.

- Low-resolution warning: if a variable has data in ≤3 distinct hours of the day, the plot still renders but warns that the observation type may not have diurnal resolution. A polar-orbiting sensor like TROPOMI samples each location roughly once per day, so a "diurnal cycle" from it is misleading. This is a warning rather than an error so sparse-overpass runs still complete.

- `kwargs["plot_dict"]` → `kwargs.get("plot_dict") or {}`, so a missing `plot_dict` no longer raises `KeyError`.

- The subplot title is only set when `domain_name` is provided, instead of rendering the literal string `None`.

## Figure hygiene

`make_boxplot` and `make_multi_boxplot` close their figures after saving; long multi-variable jobs accumulated open figures and hit matplotlib's open-figure warning.

## Dependencies

`uxarray` is optional, imported with the same `try/except ImportError` guard used for `windrose` in `surfplots.py`. The `render_unstructured_field` imports are lazy (inside the two branches), so the module imports fine without it.

> **Depends on #454 for `uxarray_render`.**


L12D2:
Fixed the unstructured grid fallback to read mio_mesh_file (what _model.py actually sets) instead of mio_grid_file, which nothing writes — MPAS meshes would previously have failed to resolve.

PR #462, opened 2026-07-29
B5c surfplots

L12D2:
# Surface plot enhancements and unstructured rendering

## What's changed

- **Wind barbs on unstructured meshes:** `wind_barb`, `u_comp`/`v_comp`, `wind_barb_step` (thinning), `wind_barb_kwargs`. Converted to knots; always uses the mean rather than percentiles. The existing barb path strides 2-D arrays, which fails for a 1-D n_face field. Added a branch that flattens lon/lat/u/v to 1-D and subsamples by wind_barb_step, plumbed uxgrid through, and extended wind_barb support to the remaining spatial plot functions.When I needed these, I had already forked develop before #434 was ingested. So, this may merge conflict or something. 

- **Hour-of-day filtering in the reduction options:** `hour_range` with `hour_basis` (solar default, or `utc`/`local`), so a reduction can be restricted to e.g. afternoon overpass hours. Solar basis offsets by longitude.

- **Unstructured rendering:** the `mio_has_unstructured_grid` branch now calls `uxarray_render.render_unstructured_field` instead of `Plot_2D`, so model fields draw on their native mesh. `plots/Plot_2D.py` is removed—after the swap nothing imports it. This is where the area average weight may need to be considered? 

- **Colorbar sizing/placement fixes** and **figure-close hygiene**.

## Dependencies

Builds on **#452** (keeps the `regulatory_tools` re-export intact).

> **Depends on #454 for `uxarray_render`.**


L12D2:
Fixed the unstructured grid fallback to read mio_mesh_file (what _model.py actually sets) instead of mio_grid_file, which nothing writes — MPAS meshes would previously have failed to resolve.

PR #463, opened 2026-07-29
Add along-track model pairing for flight-track output

L12D2:
# Along-track model pairing

Adds support for models run along an observation track (a flight path) rather than on a grid, where the model output is already 1-D in time and needs no spatial interpolation.

## `util/track_pairing.py` (new)

- `pair_track_model` — reads track model output and the obs file, resamples to a common cadence (`resample`, default 60s), and joins on time to produce a paired dataframe.

- `build_track_curtain` — builds a 2-D (time × level) curtain from track output, interpolating onto a target pressure or altitude grid.

- `save_paired_mm` — writes a paired track dataframe to netCDF in MM's layout.

- `_column_altitude` — hypsometric altitude from a pressure/temperature column.

## Driver wiring

- `is_track` on the model class, read from the model's YAML block.

- `open_models` calls only `glob_files()` for track models; `track_pairing` opens and reshapes the data during pairing, so the normal gridded open is skipped.

- `pair_data` dispatches track models to `pair_track_model` and stores the result as a standard pair object of type `aircraft`, so all existing aircraft plots work unchanged. Model variables that share an obs variable's name get the `_new` suffix the plotting code expects.

## Dependencies

Uses only `json`, `numpy`, `pandas`, and `xarray`—no new dependencies, and independent of the other PRs in this series.

## Notes

`build_track_curtain` and `_column_altitude` ship here but are not yet called from the driver. I will do the curtain-plot integration separately.

### DIFF
--- a/melodies_monet/driver/_analysis.py
+++ b/melodies_monet/driver/_analysis.py
@@ -297,6 +297,8 @@ class analysis:
                 # set the model label in the dictionary and model class instance
                 if "is_global" in self.control_dict["model"][mod].keys():
                     m.is_global = self.control_dict["model"][mod]["is_global"]
+                if "is_track" in self.control_dict["model"][mod].keys():
+                    m.is_track = self.control_dict["model"][mod]["is_track"]
                 if "mod_to_overpass" in self.control_dict["model"][mod].keys():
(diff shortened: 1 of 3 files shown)


L12D2:

This is NOT intended for v1.1

PR #464, opened 2026-07-29
Add obs2obs util for multi-observation and multi-model comparison

L12D2:
# obs2obs: compare multiple observations simultaneously

MM pairing is one-obs to N-models, so comparing two observation datasets against each other (or several obs and several models on one axis) currently means running separately and combining by hand. This util works after pairing, on the saved pair objects (`an.paired`), treating each `(pair label, variable)` as an independent data series. I hve deliberately left this as a seperate, standalone PR & util file. 

Since satellite obs-grid pairs share a lat/lon grid and model-space pairs share the model mesh, multi-obs comparison is pure alignment — no new pairing needed.

## Usage

Not wired into `_analysis.py` — it is driven explicitly after `read_analysis()`, so it adds no cost or risk to existing workflows:

```python
an.read_control()
an.read_analysis()

from melodies_monet.util import obs2obs_util

obs2obs_util.run(
    an.paired,
    an.control_dict["obs2obs"],
    default_outdir=an.output_dir,
)
```

## Series modes

Each entry in a group's `series` list is one of:

- `value` — raw values (only mix series with the same units)
- `bias` — `model - obs` from within one pair
- `relative_bias` — `100*(model-obs)/obs`, dimensionless, so surface ppb, satellite columns, and aircraft data can share one axis

## Time matching

- `none`
- `daily`
- `{window_local: [h0, h1]}`

`window_local` uses solar local hour (≈ UTC + lon/15), which is the honest way to co-locate a geostationary full-day product against a ~13:30 LT polar orbiter rather than comparing daily means with different sampling.

## Plots

- `multi_timeseries`
- `multi_boxplot`
- `diff_map`
- `scatter`
- `taylor`
- `diurnal`
- Cross-platform comparisons:
  - `xplatform_scatter`
  - `norm_spatial`
  - `operator_effect`

Where possible these route through the existing `surfplots` / `xarray_plots` functions rather than reimplementing, to reduce upkeep. It calls only:

- `make_boxplot`
- `make_scatter_density_plot`
- `make_timeseries`
- `make_diurnal_cycle`
- `make_spatial_bias_gridded`
- `make_taylor`

all of which are already on `develop`.

## Compatibility

Handles both:

- xarray pair objects (satellite: `(time, y, x)`, `(time, n_face)`, or `(time,)`)
- pandas `DataFrame`s (surface/aircraft point observations)

## Dependencies

Fully standalone — no dependency on the other PRs in this series, no new package dependencies, and nothing in the existing code path changes.

### DIFF
--- a/melodies_monet/util/obs2obs_util.py
+++ b/melodies_monet/util/obs2obs_util.py
@@ -1,0 +1,1141 @@
+"""
+Need a way to compare different obs simultaneously rather than one at a time. Because of how involved this may become, i wrote a septerate 
+obs2obs util folder to keep any issues with it seperate (as much as possible) from the rest of the MM 
+
+Goal: Use this file to put diverse obs in mm plotting compliant format using the saved paired dfs. As many plots as possible should route throug
+normal satplots or sfcplots to reduce upkeep
+
+obs2obs: compare multiple obs AND multiple models simultaneously.
+
+MELODIES-MONET pairing is one-obs to N-models; this util works AFTER pairing,
+on the saved pair objects (an.paired), treating each (pair label, variable) as
+an independent *data series*. Because the satellite obs-grid pairs share one
+lat/lon grid (and model-space pairs share the model mesh)
+
+multi-obs comparison should be pure alignment -- no new pairing needed 
+
+Series spec (each entry in a group's ``series`` list):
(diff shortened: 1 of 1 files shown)


L12D2:

This is NOT intended for v1.1

PR #465, opened 2026-07-29
B6b curtain vertcoord

L12D2:
# Curtain plot vertical coordinate and track-model integration

Wires the driver side of curtain plotting to support altitude as an alternative vertical coordinate, and to build curtains from along-track model output.

## What's changed

- `vert_coord` (new curtain group option, `pressure` default or `altitude`). With altitude, the model column is converted hypsometrically from pressure + temperature via `track_pairing._column_altitude`, and stratification happens in altitude space. If `temperature_k` isn't available the code warns and falls back to pressure rather than failing.

- Along-track models dispatch to `track_pairing.build_track_curtain`, which builds the `(time × level)` curtain directly from track output instead of going through the nearest-neighbour gridded path. Model temperature/surface variables are configurable (`model_temp_var`, `model_phis_var`, `model_ps_var`).

- Unstructured models are detected (`n_face` dim or `mio_scrip_file` attr) so the curtain path works for CESM-SE as well as gridded output.

- `vert_coord` is passed through to `airplots.make_curtain_plot`, which handles the axis direction and labelling.

- The whole curtain block in the plotting loop is replaced as one unit—it's contiguous and the surrounding `elif plot_type` branches are untouched.

## Dependencies

> **Depends on #460 (B5a, for `make_curtain_plot(vert_coord=)`) and the B6a track-pairing PR (for `build_track_curtain` / `_column_altitude`).**

No new package dependencies—`scipy.spatial.cKDTree` is already used in four other modules.


L12D2:

This is NOT intended for v1.1

